PR #23, opened 2022-03-17
feat: Replace sqlboiler with Gorm

### LINKED ISSUE #17: Switch from sqlboiler to gorm

Anondo:
<h2>resolved #17 </h2>

1. Introduced *gorm* in `db init` where connection is being established
2. Replaced *sqlboiler* generated function usage with a new `repository` package to execute orm queries using *gorm*


.gitignore
```diff
@@ -2,3 +2,5 @@
 .vscode
 _DS_Store
 .bin/
+vendor
```
moshloop, line 5: Why do we have a vendor directory ?

db/repository/config_item.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// UpdateAllFieldsConfigItem updates all the fields of a given config item row
+func (d *DBRepo) UpdateAllFieldsConfigItem(ci *models.ConfigItem) error {
+
+	ci.UpdatedAt = time.Now().UTC()
```
moshloop, line 48: Are CreatedAt and UpdatedAt not automatically catered for with Gorm ?

db/repository/config_item.go
```diff
@@ -0,0 +1,67 @@
...
+
+// NewRepo is the factory function for the database repo instance
+func NewRepo(db *gorm.DB) Database {
+	return &DBRepo{
+		db: db,
+	}
+}
+
+// GetOneConfigItem returns a single config item result
+func (d *DBRepo) GetOneConfigItem(extID string) (*models.ConfigItem, error) {
```
moshloop, line 23: What about GetConfigItem and GetConfigItems ?

moshloop:
Can we also use conventional commits ?

.gitignore
Anondo: i use `go mod vendor` to keep the dependencies in a vendor file, thats why i have added it in the `.gitignore` as not everyone uses it

db/repository/config_item.go
Anondo: Yes they are, but with the local time if i remember correctly, thought using the UTC +0 time as a default would be safe in general, hence doing it manually, will re-check and come back to this

.gitignore
moshloop: What is the advantage of that over standard global go modules ?

Anondo: Its more of a personal preference really. I like the dependencies to be isolated within the project i am working on rather than having all of them in the global mod cache with different versions. Sometimes i like to visit the source code of the dependencies and tweak here and there(temporarily speaking of course) to see different outcomes.

db/repository/config_item.go
Anondo: So a `NowFunc` (function that returns the time object to be set during created_at & updated_at) field can be set in gorm.Config during the initialization, which would determine the value automatically according to the need. Did just that.

PR #25, opened 2022-03-31
feat(scrapper): add new file scrapper

### LINKED ISSUE #19: [YAML/JSON] File Scraper
For each file found in the globbing pattern, convert it to JSON and extract the id and type

```yaml
- file:
       type: $.kind
       id: $.metadata.name
       path: 
          - config/*.yaml
          - config/*.json 
       
```

Anondo:
Introduce a new scrapper which is a json file scrapper to scrape json configuration and store it in DB. Thus resolved #19


api/v1/types.go
```diff
@@ -1,10 +1,13 @@
 package v1
 
+// ConfigScraper ...
 type ConfigScraper struct {
 	Schedule string `json:"schedule,omitempty"`
 	AWS      []AWS  `json:"aws,omitempty" yaml:"aws,omitempty"`
+	File     string `json:"file,omitempty" yaml:"file,omitempty"`
```
moshloop, line 7: The file object should be []Fille where while is 

```
Id string
Type string
Glob []string
```

cmd/config.go
```diff
@@ -89,6 +90,13 @@ func ParseConfig(configfile string, datafile string) ([]v1.ConfigScraper, error)
 	}
 
 	var scrapers []v1.ConfigScraper
+
+	if glob.Glob("config/*.json", configfile) {
```
moshloop, line 94: The globs are in the file scraper config itself and processed inside the scrapper.Scrape

scrapers/common.go
```diff
@@ -3,8 +3,10 @@ package scrapers
 import (
 	v1 "github.com/flanksource/confighub/api/v1"
 	"github.com/flanksource/confighub/scrapers/aws"
+	"github.com/flanksource/confighub/scrapers/file"
 )
 
 var All = []v1.Scraper{
 	aws.AWSScraper{},
+	file.FScrapper{},
```
moshloop, line 11:
```suggestion
	file.FileScrapper{},
```

scrapers/file/file.go
```diff
@@ -0,0 +1,33 @@
...
+
+// FScrapper ...
+type FScrapper struct {
+}
+
+// Scrape ...
+func (file FScrapper) Scrape(ctx v1.ScrapeContext, config v1.ConfigScraper) []v1.ScrapeResult {
+
+	results := []v1.ScrapeResult{}
+
```
moshloop, line 17: need to range over config.File, and then find files matching the globs, extract the id and kind, and then create a new scrape result with the contents of the file as the Config

scrapers/common.go
Anondo: Followed through this one initially, but according to the linter, using type names like <name>.<same name>* is a bad practice, in our case the sme names being the file `file.File*`. This can be ignored though

api/v1/types.go
Anondo: Possibly resolved

cmd/config.go
Anondo: Possibly resolved

scrapers/file/file.go
Anondo: Possibly resolved

scrapers/file/file.go
```diff
@@ -0,0 +1,64 @@
...
+			contentByte, err := os.ReadFile(match)
+			if err != nil {
+				logger.Errorf("failed to reading matched file: %v", err)
+				continue
+			}
+
+			jsonContent := string(contentByte)
+
+			instanceID := gjson.Get(jsonContent, "Config.InstanceId")
+			instanceType := gjson.Get(jsonContent, "Config.InstanceType")
```
moshloop, line 47:
```suggestion
			id := gjson.Get(jsonContent, fileConfig.ID)
			type := gjson.Get(jsonContent, fileConfig.Type)
```

scrapers/file/file.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, fileConfig := range config.File {
+
+		logger.Infof("Scraping JSON file id=%s type=%s", fileConfig.ID, fileConfig.Type)
+
+		globMatches := []string{}
+
+		for _, path := range fileConfig.Glob {
+			matches, err := filepath.Glob(path)
+			if err != nil {
+				logger.Errorf("failed to match glob patter: %v", err)
```
moshloop, line 30: I dont think this is an error condition - Maybe change it to a trace error, and include the glob being used.

scrapers/common.go
moshloop: it might make sense to rename this to JSONScraper, with other scrapers being added in the future e.g.  DependenciesScraper, XMLScraper, etc..

moshloop:
@Anondo - looks good, could we try adding an integration test ?

Anondo:
> @Anondo - looks good, could we try adding an integration test ?

Sure, so just to confirm, this would include just the work flow of the json scrapper right up to the point where it is saved to db ?

moshloop:
> > @Anondo - looks good, could we try adding an integration test ?
> 
> Sure, so just to confirm, this would include just the work flow of the json scrapper right up to the point where it is saved to db ?

correct - intialize a scraper with a configuration, then run it and verify its output

PR #53, opened 2022-05-30
feat: add support for github in file scrapping

Kaitou786:


db/migrations/2_seed.sql
```diff
@@ -25,6 +25,7 @@ CREATE TABLE config_item (
   zone text,
   network text,
   subnet text,
+  source text,
```
moshloop, line 28: This needs to come out, as the alter statement in 6 would then fail,

moshloop (changes requested):

Issue #49, opened 2022-05-23
iCal changes


moshloop:

too-rusty:
TODO: events are picked between now and 1 year, change it to take the input as parameters in yaml file.

PR #68, opened 2022-08-27
Ical changes 49

too-rusty:
iCal changes #49 

added ical implementation.

### DIFF
--- a/api/v1/interface.go
+++ b/api/v1/interface.go
@@ -123,6 +123,11 @@ func (s ScrapeResult) Success(config interface{}) ScrapeResult {
 	return s
 }
 
+func (s ScrapeResult) ChResult(c *ChangeResult) ScrapeResult {
+	s.ChangeResult = c
+	return s
+}
+
 func (s ScrapeResult) Errorf(msg string, args ...interface{}) ScrapeResult {
 	s.Error = fmt.Errorf(msg, args...)
 	return s
--- a/api/v1/interface.go
+++ b/api/v1/interface.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/flanksource/commons/logger"
 	fs "github.com/flanksource/confighub/filesystem"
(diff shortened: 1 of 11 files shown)


api/v1/interface.go
```diff
@@ -61,7 +62,8 @@ type AnalysisResults []AnalysisResult
 
 // Manager ...
 type Manager struct {
-	Finder fs.Finder
+	Finder    fs.Finder
+	Requester httprequest.Requester
```
moshloop, line 66: I don't think we need to use a manager for this, we are moving away from the manager object

api/v1/interface.go
```diff
@@ -121,6 +123,11 @@ func (s ScrapeResult) Success(config interface{}) ScrapeResult {
 	return s
 }
 
+func (s ScrapeResult) ChResult(c *ChangeResult) ScrapeResult {
```
moshloop, line 126: We don't use abbreviations

scrapers/ical/ical.go
```diff
@@ -0,0 +1,77 @@
...
+		results = append(results, result.Errorf("failed to parse events: %s", err.Error()))
+		return
+	}
+
+	events := make([]v1.Event, 0)
+	for _, e := range parsed {
+		events = append(events, transform(e))
+	}
+
+	result = result.
```
moshloop, line 36: So the config item itself would be the calendar with perhaps the URL and the Calendar name / Timezone, And then every event would be emitted as a Change result - the `CreatedAt` would be set to time of the event, `change_type` would be 'event', and them then description, location, status, would go into `details`

too-rusty: done, currently config item has URL only.

api/v1/interface.go
too-rusty: removed.

api/v1/interface.go
too-rusty: Removed.

PR #82, opened 2022-10-12
feat: add kubernetes scrapper

Kaitou786:


scrapers/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,53 @@
...
+
+		for _, obj := range objs {
+			createdAt := obj.GetCreationTimestamp().Time
+			results = append(results, v1.ScrapeResult{
+				BaseScraper: config.BaseScraper,
+				Name:        obj.GetName(),
+				Namespace:   obj.GetNamespace(),
+				Type:        obj.GetKind(),
+				CreatedAt:   &createdAt,
+				Config:      *obj,
```
Kaitou786, line 30: should we be persisting the entire obj or just the obj.Spec?

cc @moshloop / @Kaitou786

moshloop: I think the entire object

fixtures/kubernetes.yaml
```diff
@@ -0,0 +1,2 @@
+kubernetes:
+  -
```
moshloop, line 2:
```suggestion
  - exclusions:
       - Secret
       - ReplicaSet
       - endpoints.discovery.k8s.io
       - endpointslice.discovery.k8s.io
       - lease.coordination.k8s.io
       - podmetrics.metrics.k8s.io
       - nodemetrics.metrics.k8s.io
       - customresourcedefinition
       - controllerrevision
       - certificaterequest
       - order.acme.cert-manager.io
       
       
```

api/v1/kubernetes.go
```diff
@@ -0,0 +1,17 @@
+package v1
+
+import "github.com/flanksource/kommons"
+
+type Kubernetes struct {
+	BaseScraper     `json:",inline"`
+	Namespace       string          `json:"namespace,omitempty"`
+	UseCache        bool            `json:"useCache,omitempty"`
```
yashmehrotra, line 8: what does this field do?

Kaitou786: If set to false will invalidate the Cache for the client... so the objects are fetched every time.
See: https://github.com/flanksource/ketall/blob/master/client/client.go#L80

PR #86, opened 2022-10-13
fix: make sure release pipeline runs if semantic releaser fails

yashmehrotra:


.github/workflows/release.yml
```diff
@@ -63,19 +63,27 @@ jobs:
...
       - name: Download yq
         run: |
           wget -nv -nc -O yq https://github.com/mikefarah/yq/releases/download/v4.20.2/yq_linux_amd64
           chmod +x yq
+      - name: Set version
+        run: |
+          if ${{ needs.semantic-release.outputs.new-release-published }}; then
+            echo "RELEASE_VERSION=${{ needs.semantic-release.outputs.release-version }}" >> $GITHUB_ENV
+          else
+            export VERSION=$(git describe --abbrev=0 --tags | sed -e 's/^v//')
```
Kaitou786, line 78: I don't think we need to remove the `v` looking at the past releases... the v is present

yashmehrotra: We have to since the `version` and `appVersion` in the https://github.com/flanksource/config-db/blob/main/chart/Chart.yaml are without `v`

Kaitou786: I see okay

Kaitou786 (approved):

PR #101, opened 2022-10-26
feat: add parent_ids and paths for config_items

yashmehrotra:


db/init.go
```diff
@@ -25,6 +26,7 @@ var (
 	LogLevel         = "info"
 	HTTPEndpoint     = "http://localhost:8080/db"
 	db               *gorm.DB
+	cacheStore       *cache.Cache
```
moshloop, line 29: shouldn't this be in cache.go ?

db/update.go
```diff
@@ -39,6 +40,59 @@ func deleteChangeHandler(ctx *v1.ScrapeContext, change v1.ChangeResult) error {
 	return nil
 }
 
+func getCIParentIDFromExternal(parentExternalType, parentExternalID string) *string {
```
moshloop, line 43: I think we should maybe create an ExternalID type so that we don't have to keep replicating these arguments, and then if we add a tenant id / namespace in the future it will be easier.

I think we should also encapsulate this logicbetter e.g.  

`FindConfigItemID(external ExternalID) (*uuid.UUID, error)`

And then also make a specific select on the id and not a full GetConfigItem which has a lot of data.

db/init.go
```diff
@@ -105,6 +107,9 @@ func Init(connection string) error {
 	}
 
 	db = gormDB
+
+	// initialize cache
+	cacheStore = cache.New(1*time.Hour, 6*time.Hour)
```
moshloop, line 112: We probably. need multiple caches with different expirations:

External Type + ID to UUID => 24h

db/models/config_item.go
```diff
@@ -37,3 +39,15 @@ type ConfigItem struct {
 func (ci ConfigItem) String() string {
 	return fmt.Sprintf("%s/%s", ci.ConfigType, ci.ID)
 }
+
+type CIExternalUID struct {
```
moshloop, line 43:
```suggestion
type ExternalID struct {
```

db/config.go
```diff
@@ -25,6 +26,36 @@ func GetConfigItem(extType, extID string) (*models.ConfigItem, error) {
 	return &ci, nil
 }
 
+// GetConfigItemFromID returns a single config item result
+func GetConfigItemFromID(id string, withConfig bool) (*models.ConfigItem, error) {
```
moshloop, line 30: I would split this into 2 functions - the FromID can be removed as that would be the assumed default

GetConfigItem
GetConfigItemWithConfig

yashmehrotra: We already have a function `GetConfigItem` on line 16 of the same file which takes external type and id as parameters

moshloop: Ok but the function with a boolean parameters is still very unreadable - its better 2 create separate functions

yashmehrotra: Cool, have removed the withConfig parameter for now, we currently don't have a need for getting a single config item with config json so will create the other function when required

PR #118, opened 2022-11-21
fix: handle nil pointers for scraper

yashmehrotra:


scrapers/aws/aws.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/ssm"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/aws/smithy-go/ptr"
```
Kaitou786, line 21: Should we be bringing the entire dependency chain for ptr.ToString?

If it is the only function shouldn't we just copy it?

Kaitou786 (approved):

PR #131, opened 2023-01-06
First implementation of input field masking

adityathebe:


fixtures/file-script.yaml
```diff
@@ -3,6 +3,15 @@ file:
     id: $.id
     name: $.name
     transform:
+      mask:
```
moshloop, line 6: Can you split this example out into a different fixture and then also add an expectation for it ?

adityathebe: I've added it.

PR #132, opened 2023-01-08
feat: job history

yashmehrotra:


db/models/job_history.go
```diff
@@ -0,0 +1,38 @@
...
+	Name         string
+	SuccessCount int
+	ErrorCount   int
+	Hostname     string
+	TimeTakenMs  int64
+	ResourceType string
+	ResourceID   string
+	Details      types.JSONMap
+	TimeStart    time.Time `gorm:"-"`
+	TimeEnd      time.Time `gorm:"-"`
```
moshloop, line 20: I think we still want to record start and end, - we probably also need to a status

scrapers/runscrapers.go
```diff
@@ -36,18 +43,34 @@ func Run(ctx *v1.ScrapeContext, configs ...v1.ConfigScraper) ([]v1.ScrapeResult,
...
+					jobHistory.SuccessCount += 1
+				}
+			}
+			jobHistory.TimeEnd = time.Now()
+			if jobHistory.SuccessCount+jobHistory.ErrorCount > 0 {
+				histories = append(histories, jobHistory)
 			}
 		}
 	}
+	if err := db.SaveJobHistories(histories); err != nil {
```
moshloop, line 72: I think we should save these 1 by 1, not in bulk, and using a defer statement - We should also insert the record as th job starts with a status = running. 

It might make sense to create a JobWrapper and put it into https://github.com/flanksource/commons-db so that we can reuse across projects.

PR #133, opened 2023-01-09
Replaced otto with goja

adityathebe:
As for importing external libraries, I don't think there's a global registry in `goja` like `otto` does.
That's why `LoadSharedLibrary` now takes a vm as an argument. 

I'm not sure when the `LoadSharedLibrary` in `config-db/scrapers/processors/script.go` is called. I couldn't find any reference to it in the code. So I am not sure if it's necessary that the loaded external script is available globally.

However, to test I called LoadSharedLibrary with an eternal script after initiating a goja VM.


moshloop:
We might not having any shared libraries here, but the next step is to swop out otto with goja in canary-checker, and there we do have shared libraries: https://github.com/flanksource/canary-checker/blob/master/templating/js/js.go

PR #140, opened 2023-01-19
Replaced github.com/antonmedv/expr with github.com/google/cel-go

adityathebe:


adityathebe:
@moshloop Is there any other flanksource project that's currently using `cel-go`? 

I'm not sure if there's a better way to add operator overloads in `cel-go` because the way I've found is very verbose unlike how it's with antonmedv/expr https://github.com/flanksource/commons/blob/master/text/expressions.go#L36-L60

moshloop:
@adityathebe so the date/time functions shouldn't be needed as they are supported natively in go-cel,  for the others maybe we can create a mini app that parses the AST for functions in here: https://github.com/flanksource/gomplate/tree/main/funcs and generates a .go file for the cel functions ?

adityathebe:
@moshloop Not sure about that. Haven't worked with ASTs before. But I'm on it.

adityathebe:
Closing this PR. We're not replacing `antonmedv/expr` at the moment.

Issue #150, opened 2023-02-17
Allow configs to return events and/or cost


moshloop:
i.e. 

```yaml
full: true
```

with return expected:

```yaml
config: ...
changes: ...
cost: ...
```

PR #156, opened 2023-02-27
extract changes

adityathebe:
https://github.com/flanksource/config-db/issues/150


scrapers/file/file.go
```diff
@@ -114,9 +122,30 @@ func (file FileScraper) Scrape(ctx *v1.ScrapeContext, configs v1.ConfigScraper)
 			} else {
 				jsonContent = string(contentByte)
 			}
+
+			if config.Full {
```
moshloop, line 126: This needs to be handled in runscrapers, after any transformation have run - Any scraper should be able to return changes,

scrapers/file/file.go
```diff
@@ -154,3 +186,12 @@ func isYaml(filename string) bool {
 func isJson(filename string) bool {
 	return filepath.Ext(filename) == ".json"
 }
+
+func getConfigItemID(externalID, externalType string) (string, error) {
```
moshloop, line 190: I think there is a duty method for this

api/v1/interface.go
```diff
@@ -38,19 +38,24 @@ type AnalysisResult struct {
...
-	ExternalChangeID string
-	Action           ChangeAction
-	ChangeType       string
-	Patches          string
-	Summary          string
-	Severity         string
-	Source           string
-	CreatedAt        *time.Time
-	Details          map[string]interface{}
+	ConfigItemID     string
```
adityathebe, line 47: Is it okay to add this here?

api/v1/interface.go
```diff
@@ -131,7 +132,7 @@ type ScrapeResult struct {
 	BaseScraper         BaseScraper         `json:"-"`
 	Error               error               `json:"-"`
 	AnalysisResult      *AnalysisResult     `json:"analysis,omitempty"`
-	Changes             []ChangeResult      `json:"-"`
+	Changes             []ChangeResult      `json:"changes,omitempty"`
```
adityathebe, line 135: @moshloop not sure if this is going to cause any issue but it's needed for the test.

api/v1/interface.go
moshloop: What is the purpose of this? None of the scrapers would have this id as its generated at the db

api/v1/interface.go
moshloop: Why do we need it for the test ?

scrapers/runscrapers.go
```diff
@@ -55,19 +59,66 @@ func Run(ctx *v1.ScrapeContext, configs ...v1.ConfigScraper) ([]v1.ScrapeResult,
...
 	return results, nil
 }
+
+func getChangesFromConfig(conf any) (*v1.ChangeResult, error) {
+	configMap, ok := conf.(map[string]any)
+	if !ok {
+		return nil, errors.New("not a map")
+	}
+
+	changeMap, ok := configMap["change"].(map[string]any)
```
moshloop, line 108:
```suggestion
	changeMap, ok := configMap["changes"].(map[string]any)
```

Should this be a list ?

adityathebe: Changed to `changes`.

api/v1/interface.go
adityathebe: The expected scrape results in `fixtures/expected/file-full.json` needs to be unmarshalled to `v1.ScrapeResult`. And without a json struct tag that wouldn't be possible.

I can try something else if ``json:"-"`` is important.

scrapers/runscrapers.go
```diff
@@ -55,19 +59,66 @@ func Run(ctx *v1.ScrapeContext, configs ...v1.ConfigScraper) ([]v1.ScrapeResult,
...
+								continue
+							}
+
+							if changeRes.ExternalID != "" {
+								configItem, err := db.GetConfigItem(changeRes.ExternalType, changeRes.ExternalID)
+								if err != nil {
+									logger.Errorf("failed to get config item id: %v", err)
+								} else if configItem != nil {
+									changeRes.ConfigItemID = configItem.ID
+									scraped[i].Changes = append(scraped[i].Changes, *changeRes)
```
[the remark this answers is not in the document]
adityathebe: @moshloop I don't think it will. This only modifies the `Changes` field of the scraped result. `scraped[i].Config` is untouched.

api/v1/interface.go
adityathebe: The purpose of this is to know which config item this changes should be linked to. So after I extract the `changes` from the config file I would use it to store in the `config_id` column in the `config_changes` table.

It's used [here](https://github.com/flanksource/config-db/pull/156/files#diff-eaddcc08d43a533e9fb47a164848837e77ba33b8348431775669a878cad31d73R79)

Although, I now see that it's already handled here https://github.com/flanksource/config-db/blob/main/db/update.go#L129

adityathebe: @moshloop I removed it since it's already handled [here](https://github.com/flanksource/config-db/blob/main/db/update.go#L129)

api/v1/interface.go
adityathebe: Will revert this. Won't use expected files for the test. Will test against the db directly.

scrapers/runscrapers.go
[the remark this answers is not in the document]
adityathebe: config is extracted from the provided config.

PR #158, opened 2023-03-02
fix: config changes not being stored

adityathebe:


db/utils.go
```diff
@@ -0,0 +1,16 @@
...
+	"errors"
+
+	"github.com/jackc/pgx/v5/pgconn"
+)
+
+const (
+	pgUniqueViolationErrCode = "23505" // https://hackage.haskell.org/package/postgresql-error-codes-1.0.1/docs/PostgreSQL-ErrorCodes.html#t:ErrorCode
+)
+
+func IsUniqueConstraintPGErr(err error) bool {
```
adityathebe, line 13: gorm doesn't have a native way to identify this error https://github.com/go-gorm/gorm/issues/4037

db/utils.go
```diff
@@ -0,0 +1,16 @@
...
+	pgErr := new(pgconn.PgError)
```
moshloop, line 14:
```suggestion
        // gorm doesn't have a native way to identify this error https://github.com/go-gorm/gorm/issues/4037
	pgErr := new(pgconn.PgError)
```

db/utils.go
moshloop: nit: the details of why something what implemented in a certain way should probably go inside the function, rather on the external function docs

PR #164, opened 2023-03-03
store Kubernetes events as changes.

adityathebe:
Closes https://github.com/flanksource/config-db/issues/46


scrapers/kubernetes/events.go
```diff
@@ -0,0 +1,69 @@
...
+	if !ok {
+		return nil
+	}
+
+	reason := obj.Object["reason"].(string)
+
+	return &v1.ChangeResult{
+		ChangeType:       reason,
+		CreatedAt:        &eventCreatedAt,
+		Details:          involvedObject,
```
moshloop, line 61: The involved object isn't really relevant - it is more the full object, minus the involved object

scrapers/kubernetes/events.go
```diff
@@ -0,0 +1,94 @@
...
+	if !ok {
+		host = "<unknown-host>"
+	}
+
+	component, ok := val["component"]
+	if !ok {
+		component = "<unknown-component>"
+	}
+
+	return fmt.Sprintf("kubernetes/%s/%s", host, component)
```
adityathebe, line 56: @moshloop is this alright? When I tested this some event objects had `null` host but the component's always there. Example:
```

kubernetes/<unknown-host>/default-scheduler
kubernetes/kind-control-plane/kubelet
```

moshloop: I think maybe convert the map to key=val e.g `kubernetes/component=kubelet,host=kind-control-plane`

api/v1/kubernetes.go
```diff
@@ -6,6 +6,17 @@ import (
...
 )
 
+// SeverityKeywords is used to identify the severity
+// from the Kubernetes Event reason.
+type SeverityKeywords struct {
+	Warn  []string `json:"warn,omitempty"`
+	Error []string `json:"error,omitempty"`
+}
+
+type Event struct {
```
yashmehrotra, line 16: Use `KubernetesEvent` since its in the v1 package

scrapers/kubernetes/events.go
```diff
@@ -0,0 +1,94 @@
...
+
+	for _, k := range errKeywords {
+		if strings.Contains(strings.ToLower(reason), k) {
+			return "error"
+		}
+	}
+
+	for _, k := range warnKeywords {
+		if strings.Contains(strings.ToLower(reason), k) {
+			return "error"
```
yashmehrotra, line 33: shouldn't this be warn ?

scrapers/kubernetes/events.go
```diff
@@ -0,0 +1,94 @@
...
+	reason, _ := obj.Object["reason"].(string)
+	message, _ := obj.Object["message"].(string)
+
+	return &v1.ChangeResult{
+		ChangeType:       reason,
+		CreatedAt:        &eventCreatedAt,
+		Details:          getDetailsFromEvent(obj),
+		ExternalChangeID: string(obj.GetUID()),
+		ExternalID:       involvedObject["uid"].(string),
+		ExternalType:     ExternalTypePrefix + involvedObject["kind"].(string),
```
yashmehrotra, line 89: Should we type assert these and return err if it fails ?

yashmehrotra (changes requested):

scrapers/kubernetes/events.go
adityathebe: I believe `kind` and `uid` fields in `involvedObject` are always going to be a string https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#objectreference-v1-core.

I'll make changes so the program doesn't panic if they are nil.

api/v1/kubernetes.go
adityathebe: Made the change @yashmehrotra

scrapers/kubernetes/events.go
```diff
@@ -0,0 +1,81 @@
...
+	if !ok {
+		host = "<unknown-host>"
+	}
+
+	component, ok := val["component"]
+	if !ok {
+		component = "<unknown-component>"
+	}
+
+	return fmt.Sprintf("kubernetes/component=%s,host=%s", component, host)
```
moshloop, line 39: We shouldn't assume component or host  or put in unknown host or componenet - we need to convert the source map into key=value generically, sorted by key,

adityathebe: @moshloop I wasn't assuming that the source contains component & host only. I checked this reference - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#eventsource-v1-core to make sure of that.

adityathebe: I made the changes so it's generic now.

Issue #163, opened 2023-03-02
Delete stale configs


moshloop:
If a config item has not been scraped within the timeout, consider it deleted

```yaml
staleMode: delete
staleTimeout: 12h
``

PR #166, opened 2023-03-06
feat: delete stale config

yashmehrotra:


yashmehrotra:
Closes #163

moshloop:
@yashmehrotra, I think we might need to run this only for config scrapers that last run successfully with no errors and > 0 successes.

moshloop:
and/or run after a scraper run and not on a schedule

scrapers/stale.go
```diff
@@ -0,0 +1,37 @@
...
+            ((NOW() - updated_at) > INTERVAL '1 hour' * ?) AND
+            deleted_at IS NULL AND
+            scraper_id = ?`
+
+	result := db.DefaultDB().Exec(query, staleHours, scraperID)
+	if err := result.Error; err != nil {
+		return err
+	}
+
+	logger.Infof("Marked %d items as deleted", result.RowsAffected)
```
moshloop, line 35: I think only log if > 0

scrapers/stale.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return err
+	}
+	staleHours := int(staleDuration.Hours())
+
+	query := `
+        UPDATE config_items
+        SET deleted_at = NOW()
+        WHERE
+            ((NOW() - updated_at) > INTERVAL '1 hour' * ?) AND
```
moshloop, line 26: is this not going to be `1 hour * 24h` (the command line should probably be an int in minutes) with a default of 30

scrapers/run.go
```diff
@@ -26,5 +34,9 @@ func RunScraper(scraper v1.ConfigScraper) error {
 		return fmt.Errorf("failed to update db: %v", err)
 	}
 
+	if err = DeleteStaleConfigItems(id); err != nil {
```
moshloop, line 37: We should not be running this if there were no errors and at least 1 config returned - We want to delete as soon as possible, but not run the risk of deleting becausing there was a connection error or temporary failure

scrapers/run.go
```diff
@@ -15,15 +16,29 @@ func RunScraper(scraper v1.ConfigScraper) error {
...
 	}
 
-	if err = db.SaveResults(ctx, results); err != nil {
+	if dbErr = db.SaveResults(ctx, results); dbErr != nil {
 		//FIXME cache results to save to db later
-		return fmt.Errorf("failed to update db: %v", err)
+		return fmt.Errorf("failed to update db: %v", dbErr)
+	}
+
+	if scraperErr != nil && dbErr != nil && len(results) > 0 {
```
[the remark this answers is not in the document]
yashmehrotra: There could be, but the count of results should cover that

scrapers/run.go
```diff
@@ -15,15 +16,37 @@ func RunScraper(scraper v1.ConfigScraper) error {
...
+
+	// If error in any of the scrape results, don't delete old items
+	var errInResults = false
+	for _, r := range results {
+		if r.Error != nil {
+			errInResults = true
+			break
+		}
+	}
+	if scraperErr != nil && dbErr != nil && len(results) > 0 && !errInResults {
```
moshloop, line 46:
```suggestion
	if scraperErr == nil && dbErr == nil && len(results) > 0 && !errInResults {
```

PR #171, opened 2023-03-09
chore: add zone, region and account to aws tags

yashmehrotra:


scrapers/aws/aws.go
```diff
@@ -500,9 +500,12 @@ func (aws Scraper) instances(ctx *AWSContext, config v1.AWS, results *v1.ScrapeR
 			})
 
 			instance := NewInstance(i)
+			tags := instance.Tags
```
moshloop, line 503: I think we need to add tags on almost all the AWS resources

Issue #134, opened 2023-01-09
Map creation / delete data from config body


moshloop:
e.g. EC2 instance launch time

PR #175, opened 2023-03-14
feat: support extracting creation and deletion date from config body

adityathebe:
Resolves https://github.com/flanksource/config-db/issues/134

### DIFF
--- a/scrapers/processors/json.go
+++ b/scrapers/processors/json.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/flanksource/commons/logger"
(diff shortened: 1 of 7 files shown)


scrapers/processors/json.go
```diff
@@ -243,6 +258,24 @@ func (e Extract) extractAttributes(input v1.ScrapeResult) (v1.ScrapeResult, erro
 		}
 	}
 
+	if e.CreatedAt != nil {
+		createdAt, err := getTimestamp(*e.CreatedAt, input.Config, time.RFC3339)
```
adityathebe, line 262: Should we also make the time format configurable?

moshloop: Yes, I think both created / deleted should be a list and then the first field with a non-null value is used.

PR #177, opened 2023-03-15
fix: json struct tag for go templates.

adityathebe:
- added new fixture that showcases transformation using go template


fixtures/file-script-gotemplate.yaml
```diff
@@ -0,0 +1,17 @@
+file:
```
moshloop, line 1: Can we add an expectation?

adityathebe: added it

PR #184, opened 2023-03-23
feat: github actions scraper

yashmehrotra:


scrapers/processors/json.go
```diff
@@ -170,7 +170,7 @@ func (e Extract) Extract(inputs ...v1.ScrapeResult) ([]v1.ScrapeResult, error) {
 			}
 		default:
 			opts := oj.Options{OmitNil: true, Sort: true, UseTags: true}
-			parsedConfig, err = oj.ParseString(oj.JSON(v, opts))
+			parsedConfig, err = oj.ParseString(oj.JSON(v, &opts))
```
yashmehrotra, line 173: You have to pass `opts` as pointer else it will be silently ignored

PR #189, opened 2023-03-28
Azure Advisor

adityathebe:
Fixes https://github.com/flanksource/config-db/issues/75


adityathebe:
Advisor response

adityathebe:
```
> SELECT * FROM config_analysis;
+--------------------------------------+--------------------------------------+------------+-----------------------------------+------------------+----------+---------+--------+---------+----------+----------------------------+---------------+
| id                                   | config_id                            | created_by | analyzer                          | analysis_type    | severity | summary | status | message | analysis | first_observed             | last_observed |
|--------------------------------------+--------------------------------------+------------+-----------------------------------+------------------+----------+---------+--------+---------+----------+----------------------------+---------------|
| 0187287b-279c-89a6-bb68-4858595aade6 | 0187287b-2782-e61b-cddd-26f965da42eb | <null>     | Microsoft.Advisor/recommendations | HighAvailability | warning  | <null>  | <null> |         | {}       | 2023-03-28 13:49:32.449824 | <null>        |
| 0187287b-27ae-c997-8ca1-de18dc4affec | 0187287b-2779-ea1f-d956-3dc47d068fed | <null>     | Microsoft.Advisor/recommendations | HighAvailability | warning  | <null>  | <null> |         | {}       | 2023-03-28 13:49:32.4645   | <null>        |
| 0187287b-27b8-3b1f-733a-62c7596a5205 | 0187287b-2770-5f4d-275d-27ec7e6ce6bd | <null>     | Microsoft.Advisor/recommendations | HighAvailability | warning  | <null>  | <null> |         | {}       | 2023-03-28 13:49:32.475118 | <null>        |
+--------------------------------------+--------------------------------------+------------+-----------------------------------+------------------+----------+---------+--------+---------+----------+----------------------------+---------------+
```

scrapers/azure/azure.go
```diff
@@ -72,11 +74,11 @@ func (azure Scraper) fetchDatabases() v1.ScrapeResults {
 		for _, v := range nextPage.Value {
 			results = append(results, v1.ScrapeResult{
 				BaseScraper:  azure.config.BaseScraper,
-				ID:           *v.ID,
+				ID:           strings.ToLower(deref(v.ID)),
```
adityathebe, line 77: Needed to lowercase the ID in the config item because the azure advisor recommendation only had resource id in all lowercase. Without this I wasn't able to match analysis to config items.

Example id:  `/subscriptions/0cd017bb-aa54-4611-b21f-ecf8daee0511/resourceGroups/MC_internal-prod_internal-prod_francecentral/providers/Microsoft.Network/virtualNetworks/aks-vnet-19112733`

scrapers/azure/azure.go
```diff
@@ -72,11 +74,11 @@ func (azure Scraper) fetchDatabases() v1.ScrapeResults {
...
 				Name:         *v.Name,
 				Config:       v,
 				Type:         "RelationalDatabase",
-				ExternalType: *v.Type,
+				ExternalType: strings.ToUpper(deref(v.Type)),
```
adityathebe, line 81: Needed to uppercase this as the azure advisor recommendation had the type uppercased.

Example type: `Microsoft.Network/virtualNetworks`

.gitignore
```diff
@@ -7,3 +7,7 @@ scraped/
 
 .config-db
 cover.out
+
+# Private fixtures (contains secret credentials)
+fixtures/.*yaml
+fixtures/.*yml
```
yashmehrotra, line 13: This will ignore all the fixtures, instead keep fixtures with local secrets separately

scrapers/azure/azure.go
yashmehrotra: Add this in the code comments as well

scrapers/azure/azure.go
```diff
@@ -100,11 +102,11 @@ func (azure Scraper) fetchK8s() v1.ScrapeResults {
 		for _, v := range nextPage.Value {
 			results = append(results, v1.ScrapeResult{
 				BaseScraper:  azure.config.BaseScraper,
-				ID:           *v.ID,
+				ID:           strings.ToLower(deref(v.ID)),
 				Name:         *v.Name,
 				Config:       v,
 				Type:         "KubernetesCluster",
-				ExternalType: *v.Type,
+				ExternalType: strings.ToUpper(deref(v.Type)),
```
yashmehrotra, line 109: Use `getARMID` and `getARMExternalType` functions which encapsulate toupper and tolower

scrapers/azure/advisors.go
```diff
@@ -0,0 +1,101 @@
...
+
+func getResourceID(input *armadvisor.ResourceMetadata) string {
+	if input == nil {
+		return ""
+	}
+
+	return deref(input.ResourceID)
+}
+
+func mapAnalysisType(impactLevel *armadvisor.Category) string {
```
moshloop, line 72: I think these will need to match the ones we have on the UI filter

scrapers/azure/advisors.go
```diff
@@ -0,0 +1,101 @@
...
+		return "unknown"
+	}
+
+	switch *impactLevel {
+	case armadvisor.ImpactHigh:
+		return "critical"
+	case armadvisor.ImpactMedium:
+		return "warning"
+	default:
+		return "info"
```
yashmehrotra, line 91: These should match with our severity levels in the db

.gitignore
adityathebe: This will only ignore fixtures starting with a `.`.  But ya I think it's better to move secrets out of the repo.

scrapers/azure/advisors.go
```diff
@@ -0,0 +1,114 @@
...
+		return "other"
+	}
+
+	switch *impactLevel {
+	case armadvisor.CategoryCost:
+		return "cost"
+	case armadvisor.CategoryHighAvailability:
+		return "availability"
+	case armadvisor.CategoryOperationalExcellence:
+		return "recommendation"
```
adityathebe, line 83: @moshloop I didn't find type "recommendation" in the UI but it was being used by the AWS Trusted Advisor and I think this should be added to the UI as well.

scrapers/azure/advisors.go
adityathebe: I got these from aws trusted advisor scraper which seems to be incorrect as well.

scrapers/azure/azure.go
adityathebe: That's going to be a bit of weird solution because all these `v` are of different types and they do not expose a method to get the ID or the Type. `getARMID` & `getARMExternalType` would need to type cast all the possible types which I don't think is worth it. Let me know what you think.

scrapers/azure/advisors.go
adityathebe: @moshloop I think we should have constants in `duty` or somewhere for the analysis type and the severity level.

scrapers/azure/azure.go
yashmehrotra: No, they will always take a string arg. (v.ID and v.Type will always be string)

Something like

```golang

func getARMID(id *string) {
  // Need to lowercase the ID in the config item because the azure advisory uses resource id in all lowercase, this is required to match config analysis
  return strings.ToLower(deref(id))
}
```

scrapers/azure/advisors.go
yashmehrotra: @adityathebe We already create constant severities in duty/views, just need to add those constants in a model.
Also, if you are making that change, change severities type to enum in schema

adityathebe: Created an issue for this - https://github.com/flanksource/duty/issues/83

scrapers/azure/advisors.go
```diff
@@ -0,0 +1,116 @@
...
+		return "security"
+	}
+
+	return string(*impactLevel)
+}
+
+// mapSeverity maps the advisor impact level to a severity.
+func mapSeverity(impactLevel *armadvisor.Impact) string {
+	if impactLevel == nil {
+		return "Unknown"
```
adityathebe, line 97: probably will never be nil but I didn't know what severity to put in here.

moshloop: Use info

adityathebe: I don't think `info` is a severity level.

```
+----+----------+---------+---------+
| id | name     | aliases | icon    |
|----+----------+---------+---------|
| 1  | Critical | ['P1']  | error   |
| 2  | Blocker  | ['P2']  | error   |
| 3  | High     | ['P3']  | warning |
| 4  | Medium   | ['P4']  | info    |
| 5  | Low      | ['P4']  | info    |
+----+----------+---------+---------+
```

moshloop: Ok - then use Low

scrapers/aws/trusted_advisor.go
```diff
@@ -105,6 +105,7 @@ func (aws Scraper) trustedAdvisor(ctx *AWSContext, config v1.AWS, results *v1.Sc
 			delete(metadata, "Status")
 			analysis.Message(deref(check.Description))
 			analysis.Analysis = metadata
+			analysis.Source = "AWS Trusted Advisor"
```
adityathebe, line 108: @moshloop added source

Issue #154, opened 2023-02-21
Changes - store before/after


moshloop:
To enable better UI diff's

PR #197, opened 2023-04-05
feat: store diffs instead of patches for config changes

adityathebe:
resolves https://github.com/flanksource/config-db/issues/154


db/update.go
```diff
@@ -209,26 +212,60 @@ func SaveResults(ctx *v1.ScrapeContext, results []v1.ScrapeResult) error {
 	return nil
 }
 
-func generateDiff(a, b models.ConfigItem) (*dutyModels.ConfigChange, error) {
-	patch, err := jsonpatch.CreateMergePatch([]byte(*a.Config), []byte(*b.Config))
+// normalizeJSON returns an idented json string
+func normalizeJSON(jsonStr string) (string, error) {
```
moshloop, line 216: This should also have the keys sorted alphabetically

go.mod
```diff
@@ -271,4 +272,5 @@ require (
 replace (
 	sigs.k8s.io/kustomize/api => sigs.k8s.io/kustomize/api v0.12.1
 	sigs.k8s.io/kustomize/kyaml => sigs.k8s.io/kustomize/kyaml v0.13.9
+	github.com/flanksource/duty => /home/gunners/Projects/flanksource/duty
```
moshloop, line 275:
```suggestion
```

db/update.go
```diff
@@ -209,26 +212,60 @@ func SaveResults(ctx *v1.ScrapeContext, results []v1.ScrapeResult) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	return string(jsonStrIndented), nil
+}
+
+// generateDiff calculates the diff (git style) and patches between the
+// given 2 config items and returns a ConfigChange object if there are any changes.
+func generateDiff(newConf, prev models.ConfigItem) (*dutyModels.ConfigChange, error) {
```
moshloop, line 232: Can we add a fixture/test for this ?

db/update.go
adityathebe: @moshloop the Go `"encoding/json"` package does that by default.

db/update.go
adityathebe: Added

go.mod
adityathebe: Waiting for the [related duty PR](https://github.com/flanksource/duty/pull/101) to be merged

db/update.go
```diff
@@ -209,26 +212,71 @@ func SaveResults(ctx *v1.ScrapeContext, results []v1.ScrapeResult) error {
...
+	edits := myers.ComputeEdits("", before, after)
+	if len(edits) == 0 {
+		return "", nil
+	}
+
+	diff := fmt.Sprint(gotextdiff.ToUnified("before", "after", before, edits))
+	return diff, nil
+}
+
+// generateConfigChange calculates the diff (git style) and patches between the
```
moshloop, line 255:
```suggestion
// generateJsonPatch calculates a JSON patch between 2 files
```

adityathebe: @moshloop I think the existing comment is more descriptive. It's not just calculating json patch.

Issue #151, opened 2023-02-19
Run now api


moshloop:

adityathebe:
_argument_: id of the scraper
_returns_: the summary of the scraper result

PR #200, opened 2023-04-07
added run now endpoint

adityathebe:
resolves https://github.com/flanksource/config-db/issues/151


query/run_now.go
```diff
@@ -0,0 +1,39 @@
...
+
+		return echo.NewHTTPError(http.StatusBadRequest, err.Error()) // could mean server errors as well, but there's no trivial way to find out...
+	}
+
+	configScraper, err := scraper.V1ConfigScraper()
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "failed to transform config scraper model.", err)
+	}
+
+	if err := scrapers.RunScraper(configScraper); err != nil {
```
adityathebe, line 34: I'm aware this can take a while to run but just spinning up a Goroutine might not be the best solution and I don't think we have any other robust way of handling this right now.

@moshloop is this fine for now?

query/run_now.go
```diff
@@ -0,0 +1,39 @@
+package query
```
adityathebe, line 1: @moshloop not really sure to which package I should add this new handler. I could add it to the `scrapers` package or rename this `query` package to something more generic.

api/v1/interface.go
```diff
@@ -240,6 +240,16 @@ type QueryRequest struct {
...
 }
 
+// RunNowRequest represents the request body for a run now request
+type RunNowRequest struct {
+	ScraperID string `json:"id"`
+}
+
+// RunNowResponse represents the response body for a run now request
+type RunNowResponse struct {
+	Total int `json:"total"`
```
moshloop, line 250: We need the success, fail, and error messages -

adityathebe: Example responses

```
{
  "total": 1,
  "success": 0,
  "failed": 1,
  "errors": [
    "secret can't be empty string"
  ]
}

{
  "total": 2,
  "success": 2,
  "failed": 0
}
```

query/run_now.go
moshloop: I think this can go to the scrapers package

query/run_now.go
```diff
@@ -0,0 +1,46 @@
...
+func RunNowHandler(c echo.Context) error {
+	var req v1.RunNowRequest
+	if err := c.Bind(&req); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+	}
+
+	scraper, err := db.GetScraper(req.ScraperID)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return echo.NewHTTPError(http.StatusNotFound, "scraper with that id was not found.")
```
moshloop, line 23: Use the actual id in the error message

query/run_now.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	configScraper, err := scraper.V1ConfigScraper()
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "failed to transform config scraper model.", err)
+	}
+
+	results, err := scrapers.RunScraper(configScraper)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "failed to run scraper.", err)
```
moshloop, line 36:
```suggestion
		return echo.NewHTTPError(http.StatusInternalServerError, "failed to run scraper", err)
```

scrapers/runscrapers_test.go
```diff
@@ -37,9 +37,9 @@ var _ = Describe("Scrapers test", func() {
 	Describe("Testing file fixtures", func() {
 		fixtures := []string{
 			"file-git",
-			"file-script",
-			"file-script-gotemplate",
-			"file-mask",
+			// "file-script",
```
moshloop, line 40: Why are these commented out ?

adityathebe: This was by accident.

I was specifically fixing test for `file-git` and commented out the rest but forgot to uncomment them.

Issue #208, opened 2023-04-19
Support connection resources


moshloop:
Allow using `connection://connection-name` in the following places:

- http.endpoint
- sql.connection
- aws.accessKey
- ado.pat
- EnvVar.connection (for certificate type connections e.g. Kubeconfig, GCP, etc.)

PR #209, opened 2023-04-20
feat: Support connection resources

adityathebe:
Resolves: https://github.com/flanksource/config-db/issues/208

- [x] http.endpoint
- [x] sql.connection
- [x] aws.accessKey
- [x] ado.pat
- [x] azure.connection

~~- [ ] EnvVar.connection (for certificate type connections e.g. Kubeconfig, GCP, etc.)~~


scrapers/azure/devops/client.go
```diff
@@ -115,6 +117,13 @@ func NewAzureDevopsClient(ctx *v1.ScrapeContext, ado v1.AzureDevops) (*AzureDevo
 	if err != nil {
 		return nil, err
 	}
+
+	if _connection, err := scraper.FindConnectionFromConnectionString(ctx, db.DefaultDB(), token); err != nil {
+		return nil, fmt.Errorf("failed to find connection from (token=%s): %w", token, err)
+	} else if _connection != nil {
+		token = _connection.Password
```
adityathebe, line 124: @moshloop Not sure which field the personal access token is supposed to be stored in. I've chosen `password` here.

adityathebe:
>  EnvVar.connection (for certificate type connections e.g. Kubeconfig, GCP, etc.)

I'm not sure what needs to be done with this one.

scrapers/aws/aws_session.go
```diff
@@ -85,9 +87,18 @@ func loadConfig(ctx *v1.ScrapeContext, conn v1.AWSConnection, region string) (*a
 }
 
 func getAccessAndSecretKey(ctx *v1.ScrapeContext, conn v1.AWSConnection) (string, string, error) {
-	connection, err := duty.HydrateConnection(ctx, ctx.Kubernetes, db.DefaultDB(), conn.GetModel(), ctx.GetNamespace())
+	connection := conn.GetModel()
+
+	if _connection, err := scraper.FindConnectionFromConnectionString(ctx, db.DefaultDB(), connection.Username); err != nil {
```
moshloop, line 92: Why can't we use duty.FindConnection() ?

moshloop:
> > EnvVar.connection (for certificate type connections e.g. Kubeconfig, GCP, etc.)
> 
> I'm not sure what needs to be done with this one.

e.g. For Kubernetes we need to map the kubeconfig to connection.certificate = It seems custom kubeconfig support is commented out though: https://github.com/flanksource/config-db/blob/main/scrapers/kubernetes/kubernetes.go#L191

Also need to add the azure connection which would use username/password and properties.tenant - Subscription wouldn't be part of the connection

scrapers/aws/aws_session.go
adityathebe: `scraper.FindConnectionFromConnectionString` uses `duty.FindConnection` under the hood but additionally also parses the new connection string format.

scrapers/azure/devops/client.go
moshloop: Rather rely on the error coming from duty,  the failed to find connection should only ever occur if we have a valid URL passed, otherwise assume its a secret.

PR #226, opened 2023-05-18
feat: allow serviceAccount annotation configuration

BrendanGalloway:
- allows override to point to existing SA
- allows annotation injection when creating S
- cleaned up some nindent vs indent discrepencies

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -2,7 +2,7 @@
 Expand the name of the chart.
 */}}
 {{- define "config-db.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- .Values.nameOverride | default .Chart.Name | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
 {{/*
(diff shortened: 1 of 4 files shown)


chart/templates/_helpers.tpl
```diff
@@ -12,6 +12,15 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/* Create the name of config-db service account */}}
+{{- define "config-db.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ default (include "kube-prometheus-stack.operator.fullname" .) .Values.prometheusOperator.serviceAccount.name }}
```
moshloop, line 18: I think this is a copy and paste error

moshloop (changes requested):

chart/templates/_helpers.tpl
moshloop: @BrendanGalloway ^

yashmehrotra (approved):

yashmehrotra:
@moshloop Lets merge this ?

Issue #224, opened 2023-05-17
A scraper Run needs to have a context deadline


adityathebe:
Ideally we want the context timeout to be less than the cron schedule.

adityathebe:
Similar https://github.com/flanksource/canary-checker/issues/1004

PR #228, opened 2023-05-22
feat: prevent concurrent execution of config scraper jobs

adityathebe:
Resolves: https://github.com/flanksource/config-db/issues/224

Tested locally with Trivy. Made sure only one instance of Trivy was running

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -62,7 +62,7 @@ func serve(configFiles []string) {
 func startScraperCron(configFiles []string) {
 	scraperConfigsFiles, err := v1.ParseConfigs(configFiles...)
 	if err != nil {
-		logger.Fatalf(err.Error())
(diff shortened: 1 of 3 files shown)


cmd/server.go
```diff
@@ -88,10 +88,10 @@ func startScraperCron(configFiles []string) {
 
 		fn := func() {
 			if _, err := scrapers.RunScraper(_scraper); err != nil {
-				logger.Errorf("Error running scraper: %v", err)
+				logger.Errorf("Error running scraper(id=%s): %v", _scraper.ID, err)
 			}
 		}
-		defer fn()
```
adityathebe, line 94: Not sure why the first run was done sequentially. I think running them all concurrently makes sense & the cron manager also does the same.

PR #235, opened 2023-05-24
chore: add PodMetrics and NodeMetrics to scraper exclusions

yashmehrotra:


scrapers/kubernetes/kubernetes.go
```diff
@@ -228,10 +228,19 @@ func cleanKubernetesObject(obj map[string]any) string {
 	o.Delete("metadata", "annotations", "control-plane.alpha.kubernetes.io/leader")
 	o.Delete("status", "artifact", "lastUpdateTime")
 
+	// Canaries
```
moshloop, line 231: Shouldn't this be part of the fixture as a transform: exclude ?

PR #242, opened 2023-05-29
added more delete events to config changes rule

adityathebe:
Resolves: https://github.com/flanksource/config-db/issues/149


scrapers/changes/rules.yaml
```diff
@@ -1,2 +1,734 @@
+AbortEnvironmentUpdate:
```
moshloop, line 1: Can we not add support for wildcards ?

scrapers/changes/rules.yaml
```diff
@@ -1,2 +1,16 @@
...
+Disable*:
+  action: delete
+Remove*:
+  action: delete
+Stop*:
+  action: delete
+Terminate*:
+  action: delete
+WipeDevice:
+  action: delete
```
moshloop, line 16:
```suggestion
Delete*:
  action: delete
Remove*:
  action: delete
Terminate*:
  action: delete
```

Issue #240, opened 2023-05-25
Resolve analysis items that are no longer observed


moshloop:
Matching logic for alertmanager check resolutuion

+ Removed resolved analysis from default queries
+ Provide an option to show resolved analysis items in the full listing/detail panel (same as deleted configs)

PR #243, opened 2023-05-29
feat: mark analyses as resolved if they are not observed anymore

adityathebe:
* Save scraper ID on config analysis

Resolves: https://github.com/flanksource/config-db/issues/240

Depends on: https://github.com/flanksource/duty/pull/168


db/update.go
```diff
@@ -185,14 +187,39 @@ func updateAnalysis(ctx *v1.ScrapeContext, result *v1.ScrapeResult) error {
...
 
+func GetCurrentDBTime(ctx context.Context) (time.Time, error) {
+	var latest time.Time
+	err := db.WithContext(ctx).Raw(`SELECT CURRENT_TIMESTAMP`).Scan(&latest).Error
+	return latest, err
+}
+
+// UpdateAnalysisStatusBefore updates the status of config analyses that were last observed before the specified time.
+func UpdateAnalysisStatusBefore(ctx context.Context, before time.Time, scraperID, status string) error {
+	db.Debug()
```
moshloop, line 206:
```suggestion
```

db/update.go
```diff
@@ -185,14 +187,39 @@ func updateAnalysis(ctx *v1.ScrapeContext, result *v1.ScrapeResult) error {
...
+		Error
+}
+
 // SaveResults creates or update a configuartion with config changes
 func SaveResults(ctx *v1.ScrapeContext, results []v1.ScrapeResult) error {
-	for _, result := range results {
+	startTime, err := GetCurrentDBTime(ctx)
+	if err != nil {
+		logger.Errorf("unable to get current db time: %v", err)
+	}
```
moshloop, line 220: There are 2 scenarios - 

1. DB is configured but cannot connect - this should error out and return
2. DB is not configured, in which case don't attempt to get db time in the first place.

db/update.go
```diff
@@ -185,14 +187,38 @@ func updateAnalysis(ctx *v1.ScrapeContext, result *v1.ScrapeResult) error {
...
+	if analysis.Status == "" {
+		analysis.Status = dutyModels.AnalysisStatusOpen
+	}
 
 	return CreateAnalysis(analysis)
 }
 
+func GetCurrentDBTime(ctx context.Context) (time.Time, error) {
+	var now time.Time
+	err := db.WithContext(ctx).Raw(`SELECT CURRENT_TIMESTAMP`).Scan(&now).Error
```
moshloop, line 200: @adityathebe  not sure I understand how this works without a DB connection ?

adityathebe:
image

Fixed Trivy formatting in this PR as well

db/update.go
adityathebe: "config-db needs to run without a db connection"

adityathebe: @moshloop We don't attempt to save the results in run mode if db connection string is not provided. So I think we're good

https://github.com/flanksource/config-db/blob/3bfafa23659e6847f5850f245f80fd0b78c77bd4/cmd/run.go#L40-L54

adityathebe:
@moshloop Tried out markdown parser and looks better. However, I'm not sure if all the datasources return markdown description. Debian Security Tracker seems to return markdown.


image
I've also added datasource.
image

PR #244, opened 2023-05-30
feat: add retention period for changes and analysis

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/808


moshloop:
@yashmehrotra can we update the charts as well ?

yashmehrotra:
> @yashmehrotra can we update the charts as well ?

Done

PR #250, opened 2023-06-06
fix: handle scrape config deletions

yashmehrotra:


yashmehrotra:
@moshloop @adityathebe 

How should we handle dangling config_items here ? Mark them as deleted too ?

adityathebe:
> How should we handle dangling config_items here ? Mark them as deleted too ?

A user might want to stop the scraping and still keep the scraped items ...?

moshloop:
I think deleting the scraped items makes the most sense, if you don't want to run the scraper, then we should be able to disable it instead.

db/config_scraper.go
```diff
@@ -24,11 +27,45 @@ func FindScraper(id string) (*models.ConfigScraper, error) {
...
+	selectQuery := strings.Join(selectQueryItems, " UNION ")
+
+	var referredConfigIDs []string
+	logger.Infof("YASH Reffered ids are %v", referredConfigIDs)
+	if err := db.Raw(selectQuery).Scan(&referredConfigIDs).Error; err != nil {
+		return err
+	}
+
+	// Remove scraper_id from linked config_items
+	if err := db.Table("config_items").Where("id IN (?) AND scraper_id = ?", referredConfigIDs, id).Update("scraper_id", nil).Error; err != nil {
```
moshloop, line 60: Not sure this will perform well with thousands or tens of thousands of configs - Can we not update with a subselect ?

db/config_scraper.go
```diff
@@ -24,11 +26,47 @@ func FindScraper(id string) (*models.ConfigScraper, error) {
...
+func DeleteScrapeConfig(id string) error {
+	if err := db.Table("config_scrapers").
+		Where("id = ?", id).
+		Update("deleted_at", time.Now()).
+		Error; err != nil {
+		return err
+	}
+
+	// Fetch all IDs which are linked to other tables
+	foreignKeyTables := []string{
```
moshloop, line 38: I don't think we want to retain config items with changes / analysis - this will almost always be the case - I think only for evidences

PR #254, opened 2023-06-07
fix: set default token permissions to read-only for workflows

dabasvibhor:


moshloop:
@dabasvibhor The permissions for the test workflow will need to be updated to allow it to post check runs

PR #264, opened 2023-06-15
chore: remove generation and uptime from kubernetes status

yashmehrotra:
Fixes: https://mission-control.azure.lab.flanksource.com/incidents/0188b498-865c-e51b-848e-56e508557681 & https://mission-control.azure.lab.flanksource.com/incidents/0188b497-f9c5-466a-a820-bebc3620fae0

### DIFF
--- a/fixtures/kubernetes.yaml
+++ b/fixtures/kubernetes.yaml
@@ -33,3 +33,7 @@ kubernetes:
         - jsonpath: $.status.lastCheck
         - jsonpath: $.status.lastTransitionedTime
         - jsonpath: $.status.checkStatus.*.lastTransitionedTime
+        - jsonpath: $.status.checkStatus.*.uptime1h
+        - jsonpath: $.status.checkStatus.*.latency1h
+        - jsonpath: $.status.uptime1h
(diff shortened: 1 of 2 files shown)


fixtures/kubernetes.yaml
```diff
@@ -33,3 +33,7 @@ kubernetes:
         - jsonpath: $.status.lastCheck
         - jsonpath: $.status.lastTransitionedTime
         - jsonpath: $.status.checkStatus.*.lastTransitionedTime
+        - jsonpath: $.status.checkStatus.*.uptime1h
+        - jsonpath: $.status.checkStatus.*.latency1h
+        - jsonpath: $.status.uptime1h
+        - jsonpath: $.status.latency1h
```
yashmehrotra, line 39: @moshloop Can we move all these to native  scraper ignorance implementation ?

Since, config-db will almost always consume Canary CRDs, these exclude jsonpaths need to implemented and migrated everywhere along with kubernetes spec

Since these are custom to us, we can remove them all the time, will also make default k8s spec cleaner out of the box

PR #266, opened 2023-06-15
fix: ignore config scrapers of other agents

adityathebe:


db/config_scraper.go
```diff
@@ -79,7 +79,7 @@ func PersistScrapeConfigFromCRD(scrapeConfig *v1.ScrapeConfig) (bool, error) {
 
 func GetScrapeConfigs() ([]models.ConfigScraper, error) {
 	var configScrapers []models.ConfigScraper
-	err := db.Find(&configScrapers, "deleted_at IS NULL").Error
+	err := db.Find(&configScrapers, "deleted_at IS NULL AND agent_id = ?", uuid.Nil).Error
```
moshloop, line 82: Could we add an agent-name cli argumentfor this as well (defaulting to nil)

db/config_scraper.go
```diff
@@ -77,9 +77,9 @@ func PersistScrapeConfigFromCRD(scrapeConfig *v1.ScrapeConfig) (bool, error) {
 	return tx.RowsAffected > 0, tx.Error
 }
 
-func GetScrapeConfigs() ([]models.ConfigScraper, error) {
+func GetScrapeConfigsOfAgent(agentID uuid.UUID) ([]models.ConfigScraper, error) {
 	var configScrapers []models.ConfigScraper
-	err := db.Find(&configScrapers, "deleted_at IS NULL").Error
+	err := db.Find(&configScrapers, "deleted_at IS NULL AND agent_id = ?", agentID).Error
```
moshloop, line 82: is agentID going to be `nil` or `0000-0000-000` - what is the default value in the table ?

adityathebe: The default value is `0000-0000-000`

PR #276, opened 2023-07-05
feat: add health status for kubernetes objects

yashmehrotra:
Fixes: #275


scrapers/kubernetes/kubernetes.go
```diff
@@ -116,6 +117,11 @@ func (kubernetes KubernetesScraper) Scrape(ctx *v1.ScrapeContext, configs v1.Con
 			}
 			tags["cluster"] = config.ClusterName
 
+			// Add health metadata
+			if healthStatus, err := health.GetResourceHealth(obj, nil); err == nil {
```
moshloop, line 121: Can we use the status and description db columns ?

yashmehrotra: Ok

Issue #280, opened 2023-07-08
Properties parser doesn't handle tabs/comments


yashmehrotra:
Mission control Link: https://mission-control.azure.lab.flanksource.com/incidents/01892b0e-bd53-11c3-f5c0-26ff350b4b3f

PR #281, opened 2023-07-11
fix: remove comments and tabs from properties file

yashmehrotra:
Fixes: #280


moshloop:
@yashmehrotra  can we add an example of the postgresql conf e.g. https://github.com/postgres/postgres/blob/master/src/backend/utils/misc/postgresql.conf.sample

yashmehrotra:
@moshloop Have added a test for it

PR #288, opened 2023-07-28
feat: use of CRD to run scrapers

adityathebe:
- Includes quite a few refactoring
  - utilization of scrape configs from the context
  - use of scrapeConfig CRD


fixtures/file-mask.yaml
```diff
@@ -1,16 +1,21 @@
...
-          jsonpath: $.password
-          value: md5sum
-        - selector:
-            type: Config
-          jsonpath: $.secret
-          value: '***'
-    paths:
-      - fixtures/data/single-config.json
+apiVersion: configs.flanksource.com/v1
+kind: Config
```
moshloop, line 2: I think this is ConfigScraper

Issue #279, opened 2023-07-07
Kubernetes watch based changes


moshloop:

moshloop:
e.g. 
- [ ] Add a kubernetes watch on events, to insert new changes in real-time, 
- [ ] Then for those configs the events operate on, trigger a fresh get, but using a 5 second buffer, e.g. if 5 events for the same object fire in 5 seconds we should only do an incremental get once

PR #296, opened 2023-09-05
feat: watch kubernetes events

adityathebe:
Resolves: https://github.com/flanksource/config-db/issues/279


scrapers/kubernetes/events_watch.go
```diff
@@ -0,0 +1,84 @@
...
+		case watchEvent := <-watcher.ResultChan():
+			var event Event
+			if err := event.FromObjMap(watchEvent.Object); err != nil {
+				logger.Errorf("failed to unmarshal event (id=%s): %v", event.GetUID(), err)
+				continue
+			}
+
+			logger.Infof("New Event: reason=%s source=%s", event.Reason, event.Source)
+
+			if utils.MatchItems(event.Reason, config.Event.Exclusions...) {
```
moshloop, line 55: I think this determines whether to track it as a change, but not wether to requeue another sync

scrapers/kubernetes/events_watch.go
```diff
@@ -0,0 +1,84 @@
...
+			if len(changesBuffer) >= maxBufferSize {
+				consumeChangeEvents(ctx, consume)
+			}
+		}
+	}
+}
+
+// consumeChangeEvents fetches the configs referenced by the changes and saves them.
+// It clears the buffer after.
+func consumeChangeEvents(ctx *v1.ScrapeContext, consume consumerFunc) {
```
moshloop, line 75: I don't think this will do what we want - If only a few events are fired an hour it will take too long for them to be consumed  it might be better to have a map with uid -> min(date), and then iterate over that map every 10 seconds and deleting the items as they are processed.

adityathebe: > If only a few events are fired an hour it will take too long for them to be consumed

But we have a _(10 second)_ timeout though. So no matter what, we consume whatever amount of events that's been gathered so far.

scrapers/kubernetes/events_watch.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		// TODO: Definitely need a new exclusion list in the kubernetes spec to exclude by pattern matching names (junit*, for example)
+		if strings.Contains(event.InvolvedObject.Name, "junit") || strings.Contains(event.InvolvedObject.Name, "hello-world") {
+			continue
+		}
+
+		if collections.Contains([]string{"canaries", "monitoring"}, event.InvolvedObject.Namespace) {
+			continue
+		}
```
adityathebe, line 74: @moshloop I don't think we can exclude by labels at this stage because the involvedObject doesn't contain labels. It only contains name, namespace, kind and few other data https://github.com/kubernetes/api/blob/4c0435861abb07eb6505ad0feda175c1d79b4bfc/core/v1/types.go#L6076-L6111

I think we can refactor our current config structure; instead of having a `exclusions` list that accepts name, shortnames, kind, ... we can make it more structured ...?

```yaml
apiVersion: configs.flanksource.com/v1
kind: ScrapeConfig
metadata:
  name: kubernetes-scraper
spec:
  kubernetes:
    - clusterName: local-kind-cluster
      exclusions:
        kind:
          - Secret
          - ReplicaSet
          - APIService
          - PodMetrics
          - NodeMetrics
        name:
          - junit*
          - hello-world*
        namespace:
          - canaries
          - monitoring
        labels:
          prod: dev
```

We would need to modify ketall to accepts patterns in exclusions so it's not just direct string match.

cmd/server.go
```diff
@@ -107,6 +108,11 @@ func startScraperCron(configFiles []string) {
 			}
 		}
 		go scrapers.AtomicRunner(scraper.ID.String(), fn)()
+
+		for _, k := range _scraper.Spec.Kubernetes {
+			ctx := api.NewScrapeContext(context.Background(), _scraper)
+			go exitOnError(kubernetes.WatchEvents(ctx, k, kubernetesChangeEventConsumer), "error watching events")
```
moshloop, line 114: We should not be exiting on error, only retrying with a backoff

scrapers/kubernetes/events_watch.go
```diff
@@ -0,0 +1,103 @@
...
+		if err := event.FromObjMap(watchEvent.Object); err != nil {
+			logger.Errorf("failed to unmarshal event (id=%s): %v", event.GetUID(), err)
+			continue
+		}
+
+		if event.InvolvedObject == nil {
+			continue
+		}
+
+		if config.Exclusions.Filter(event.InvolvedObject.Name, event.InvolvedObject.Namespace, event.InvolvedObject.Kind, nil) {
```
moshloop, line 64: Why not just pass `event` through?

scrapers/kubernetes/kubernetes.go
```diff
@@ -25,17 +24,65 @@ func (kubernetes KubernetesScraper) CanScrape(configs v1.ScraperSpec) bool {
...
+func DecodeItemID(encoded string) ItemID {
+	parts := strings.Split(encoded, "/")
+	return ItemID{
+		Name:      parts[0],
+		Namespace: parts[1],
+		Kind:      parts[2],
+	}
+}
+
+func (kubernetes KubernetesScraper) ScrapeSome(ctx *v1.ScrapeContext, configRaw any, ids []string) v1.ScrapeResults {
```
moshloop, line 46: Why is this a configRaw any ?

moshloop: I would use the name IncrementalScrape

adityathebe: > Why is this a configRaw any ?

```
// TargettedScraper can additionally scrape a predefined set of configs.
// +kubebuilder:object:generate=false
type TargettedScraper interface {
	Scraper
	ScrapeSome(ctx *ScrapeContext, config any, ids []string) ScrapeResults
}
```

`ScrapeSome` needs to support config types for all scrapers and there's no common interface I could find, so it's `any`.

---
Also, I had tried passing in `configIndex` since the actual config is already in the `ctx.ScrapeConfig.Spec`, but felt passing the config was more "natural".

scrapers/kubernetes/events_watch.go
adityathebe: I generally lean towards passing in the args explicitly instead of passing objects as it's not immediately clear what fields of the objects are required by the function.

moshloop: Why do we need to know what fields they function uses? that is breaking encapsulation which says we shouldn't expose the inner workings of the function

adityathebe: > Why do we need to know what fields they function uses? 

_Example_: sometimes you need to craft the object you're passing to the function _(may not apply to this particular case)_.  Or the object could come from multiple sources such that the sources do not have the required fields populated _(example: a getResource() db function that only returns name and id whereas the function needs name, id and properties)_. 

If the objects has N fields, you don't really know which of those N fields you need to populate before you pass to the function until you read the function source code.

> Why do we need to know what fields they function uses? that is breaking encapsulation which says we shouldn't expose the inner workings of the function

This isn't really exposing the inner workings though. It's just making it explicit what parameters the function requires.

moshloop: This is true for parameters, but the event object is the parameter itself, you wont create a new event object by hand to pass through to a filter

adityathebe: Ya in this particular case we might never run into this issue. I did it this way out of habit.

adityathebe: will make the change.

scrapers/kubernetes/events_watch.go
```diff
@@ -0,0 +1,103 @@
...
+
+	watcher, err := ctx.Kubernetes.CoreV1().Events(config.Namespace).Watch(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create a new event watcher: %w", err)
+	}
+	defer watcher.Stop()
+
+	for watchEvent := range watcher.ResultChan() {
+		var event Event
+		if err := event.FromObjMap(watchEvent.Object); err != nil {
```
moshloop, line 55: Why do we need todo this - why can't we just use watchEvent directly ?

scrapers/kubernetes/events_watch.go
```diff
@@ -0,0 +1,103 @@
...
+	for {
+		time.Sleep(refetchEventResourceInterval)
+
+		if len(t.involvedObjects) == 0 {
+			continue
+		}
+
+		t.lock.Lock()
+		consume(ctx, config, t.involvedObjects)
+		t.involvedObjects = make(map[string]map[string]*InvolvedObject)
```
moshloop, line 100: This will hold the lock for too long - We should keep a local rerence to the current map, and then set t.involved to a new map, then unlock, then consume

scrapers/kubernetes/events_watch.go
adityathebe: WatchEvent is an interface with very limited methods

scrapers/kubernetes/kubernetes.go
moshloop: Why do we need todo this at all? Nothing else will call Kubernetes.ScrapeIncremental but the event consumer?

cmd/server.go
```diff
@@ -130,3 +136,22 @@ func forward(e *echo.Echo, prefix string, target string) {
...
+	if err != nil {
+		logger.Fatalf("%s %v", description, err)
+	}
+}
+
+func kubernetesChangeEventConsumer(ctx *v1.ScrapeContext, config v1.Kubernetes, resourcesPerKind map[string]map[string]*kubernetes.InvolvedObject) {
+	var resourceIDs []string
+	for kind, resources := range resourcesPerKind {
+		for _, r := range resources {
+			resourceIDs = append(resourceIDs, kubernetes.ItemID{Kind: kind, Name: r.Name, Namespace: r.Namespace}.Encode())
```
moshloop, line 150: Why do we need to encode/decode the itemid - why not just passthrough itemid itself ?

I would also rename itemid to ResourceId

go.mod
```diff
@@ -39,54 +39,54 @@ require (
...
 	github.com/uber/athenadriver v1.1.14
 	github.com/xo/dburl v0.13.1
 	gopkg.in/flanksource/yaml.v3 v3.2.3
 	gopkg.in/yaml.v3 v3.0.1
-	gorm.io/gorm v1.25.0
-	k8s.io/apimachinery v0.26.4
-	k8s.io/client-go v0.26.4
-	sigs.k8s.io/controller-runtime v0.14.6
+	gorm.io/gorm v1.25.4
+	k8s.io/api v0.28.0
```
moshloop, line 71: Don't bump the kubernetes versions as upgrading can sometimes cause issues when importing into / from other projects

scrapers/kubernetes/events_watch.go
moshloop: Ok - but then we should use https://github.com/mitchellh/mapstructure - marshalling and unmarshalling json is very slow

adityathebe:
@moshloop This is ready

PR #298, opened 2023-09-22
fix: mark evicted pods as deleted

yashmehrotra:
Fixes: https://github.com/flanksource/config-db/issues/284

### DIFF
--- a/db/config.go
+++ b/db/config.go
@@ -19,7 +19,7 @@ import (
 // GetConfigItem returns a single config item result
 func GetConfigItem(extType, extID string) (*models.ConfigItem, error) {
 	ci := models.ConfigItem{}
-	tx := db.Limit(1).Find(&ci, "type = ? and external_id  @> ?", extType, pq.StringArray{extID})
+	tx := db.Limit(1).Select("id", "config_class", "type", "config").Find(&ci, "type = ? and external_id  @> ?", extType, pq.StringArray{extID})
 	if tx.RowsAffected == 0 {
 		return nil, nil
 	}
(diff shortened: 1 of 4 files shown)


db/config.go
```diff
@@ -72,14 +72,14 @@ func CreateConfigItem(ci *models.ConfigItem) error {
...
-func UpdateConfigItem(ci *models.ConfigItem) error {
+func UpdateConfigItem(ci *models.ConfigItem, updateDeletedAt bool) error {
 
 	if err := db.Updates(ci).Error; err != nil {
 		return err
 	}
 
 	// Since gorm ignores nil fields, we are setting deleted_at explicitly
-	if ci.DeletedAt != nil {
+	if updateDeletedAt {
```
moshloop, line 82: It might be better to put this on the configitem itself as ci.TouchDeletedAt or otherwise so that the api doesn't change

scrapers/kubernetes/kubernetes.go
```diff
@@ -124,6 +125,23 @@ func (kubernetes KubernetesScraper) Scrape(ctx *v1.ScrapeContext) v1.ScrapeResul
...
+			if !obj.GetDeletionTimestamp().IsZero() {
+				deletedAt = &obj.GetDeletionTimestamp().Time
+			}
+
+			// Evicted Pods must be considered deleted
+			if obj.GetKind() == "Pod" && status == string(health.HealthStatusDegraded) {
+				objStatus := obj.Object["status"].(map[string]any)
+				if val, ok := objStatus["reason"].(string); ok && val == "Evicted" {
+					// Use createdAt as default and try to parse the evict time
+					deletedAt = &createdAt
```
moshloop, line 138: It probably makes more sense to use now() as the default

yashmehrotra: If we use now and the parsing fails, the deleted time will keep on chaning

PR #300, opened 2023-09-26
feat: upstream sync

adityathebe:
Resolves: https://github.com/flanksource/config-db/issues/286

Merge after:
- [x] https://github.com/flanksource/config-db/pull/299
- [x] https://github.com/flanksource/postq/pull/1
- [x] https://github.com/flanksource/duty/pull/294


jobs/jobs.go
```diff
@@ -5,21 +5,46 @@ import (
...
+	"github.com/flanksource/config-db/api"
 	"github.com/robfig/cron/v3"
 )
 
 var FuncScheduler = cron.New()
 
+const (
+	PullConfigScrapersFromUpstreamSchedule = "@every 30s"
+	PushConfigResultsToUpstreamSchedule    = "@every 10s"
+	ReconcileConfigsToUpstreamSchedule     = "@every 3h"
```
adityathebe, line 17: @moshloop let me know what these schedules should be.

jobs/sync_upstream.go
```diff
@@ -0,0 +1,217 @@
...
+		return err
+	}
+
+	if err := ctx.DB().Where("id > ?", t.status.AnalysisID).Find(&pushData.ConfigAnalysis).Error; err != nil {
+		return err
+	}
+
+	if err := ctx.DB().Where("id > ?", t.status.ChangeID).Find(&pushData.ConfigChanges).Error; err != nil {
+		return err
+	}
```
adityathebe, line 187: @moshloop Since we're using id instead of time to push data, wouldn't this mean that updates to already pushed records are not pushed anymore?

adityathebe:
## Pulling scrape configs 

a. Pull everything that was updated since X timestamp
- X could be empty on startup so agent fetches all scrape configs
- or it could be a delta of 5 minutes.

## Push config

### A.  Primary means of pushing to upstream

  a. use event_queue to push real time changes to upstream, **OR**
  b. On a schedule X, fetch all the results that were updated in the last (X + delta) interval. _(Although config_analysis doesn't have updated_at table so we might need to add it.)_

### B. Reconcile

Every 8 hours, push all results that were updated in the last 48 hours.

jobs/sync_upstream.go
```diff
@@ -0,0 +1,217 @@
...
+	"config_items",
+	"config_changes",
+	"config_analysis",
+}
+
+// ReconcileConfigScraperResults pushes missing scrape config results to the upstream server
+func ReconcileConfigScraperResults() {
+	ctx := api.DefaultContext
+
+	jobHistory := models.NewJobHistory("PushScraperConfigResultsToUpstream", "Config", "")
```
moshloop, line 28:
```suggestion
	jobHistory := models.NewJobHistory("PushUpstream", "Config", "")
```

jobs/jobs.go
```diff
@@ -5,21 +5,39 @@ import (
...
 
 	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/config-db/api"
 	"github.com/robfig/cron/v3"
 )
 
 var FuncScheduler = cron.New()
 
+const (
+	PullConfigScrapersFromUpstreamSchedule = "@every 30s"
```
moshloop, line 15:
```suggestion
	PullConfigScrapersFromUpstreamSchedule = "@every 5m"
```

cmd/root.go
```diff
@@ -78,7 +73,22 @@ func ServerFlags(flags *pflag.FlagSet) {
...
+	var upstreamPageSizeDefault = 500
+	if val, exists := os.LookupEnv("UPSTREAM_PAGE_SIZE"); exists {
+		if parsed, err := strconv.Atoi(val); err != nil || parsed < 0 {
+			logger.Fatalf("invalid value=%s for UPSTREAM_PAGE_SIZE. Must be a postive number", val)
+		} else {
+			upstreamPageSizeDefault = parsed
+		}
+	}
+
+	flags.StringVar(&api.UpstreamConfig.Host, "upstream-host", os.Getenv("UPSTREAM_HOST"), "central mission control instance to sync scrape configs & their results")
```
moshloop, line 87: I think we need to add chart values as well

moshloop:
> ## Pulling scrape configs
> a. Pull everything that was updated since X timestamp
> 
> * X could be empty on startup so agent fetches all scrape configs
> * or it could be a delta of 5 minutes.
> 
> ## Push config
> ### A. Primary means of pushing to upstream
> a. use event_queue to push real time changes to upstream, **OR** b. On a schedule X, fetch all the results that were updated in the last (X + delta) interval. _(Although config_analysis doesn't have updated_at table so we might need to add it.)_
> 
> ### B. Reconcile
> Every 8 hours, push all results that were updated in the last 48 hours.
> 
> @moshloop wdyt?


Pulling scrape configs every 5 minutes or so is fine

For push I think the event_queue mechanism as the primary is fine + normal reconciliation. 

We will probably need to included updated_at in the hash,  and then in get upstream details return both ids and updated_at on hash mismatch, and only push the difference

chart/values.yaml
```diff
@@ -61,6 +61,12 @@ serviceAccount:
   name: ""
   annotations: {}
 
+upstream:
+  secretKeyRef:
+    name: upstream
+  agentName:
```
BrendanGalloway, line 67: should this have a default?

chart/templates/deployment.yaml
```diff
@@ -42,6 +42,9 @@ spec:
             - --disable-postgrest={{ .Values.disablePostgrest }}
             - --change-retention-days={{ .Values.configChangeRetentionDays }}
             - --analysis-retention-days={{ .Values.configAnalysisRetentionDays }}
+          envFrom:
+            - secretRef:
+              name: {{ .Values.upstream.secretKeyRef.name }}
```
BrendanGalloway, line 47:
```suggestion
            - secretRef:
                name: {{ .Values.upstream.secretKeyRef.name }}
```

chart/values.yaml
```diff
@@ -61,6 +61,12 @@ serviceAccount:
   name: ""
   annotations: {}
 
+upstream:
+  secretKeyRef:
```
BrendanGalloway, line 65: Maybe indicate which keys this secret is expected to contain

BrendanGalloway (changes requested):

chart/templates/deployment.yaml
```diff
@@ -42,6 +42,9 @@ spec:
             - --disable-postgrest={{ .Values.disablePostgrest }}
             - --change-retention-days={{ .Values.configChangeRetentionDays }}
             - --analysis-retention-days={{ .Values.configAnalysisRetentionDays }}
+          envFrom:
```
BrendanGalloway, line 45: must this secret _always_ be supplied?  if not, either wrapped this stanza with an if statement, or set the envFrom entry as optional

chart/values.yaml
adityathebe: I suppose this done simply via a comment?

BrendanGalloway (approved):

chart/values.yaml
BrendanGalloway: yes

chart/values.yaml
```diff
@@ -61,6 +61,13 @@ serviceAccount:
   name: ""
   annotations: {}
 
+upstream:
+  enabled: true
```
moshloop, line 65:
```suggestion
  enabled: false
```

PR #309, opened 2023-11-03
feat: embedded db

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/305

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -32,6 +32,8 @@ spec:
       labels: {{- include "config-db.selectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ template "config-db.serviceAccountName" . }}
+      securityContext:
+        fsGroup: 1000
       containers:
         - name: {{ include "config-db.name" . }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -1,5 +1,9 @@
 apiVersion: apps/v1
+{{- if and (eq .Values.db.external.enabled false) (eq .Values.db.embedded.persist true) }}
+kind: StatefulSet
+{{- else }}
 kind: Deployment
+{{- end }}
 metadata:
   name: {{ include "config-db.name" . }}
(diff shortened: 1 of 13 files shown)


chart/templates/deployment.yaml
```diff
@@ -1,17 +1,39 @@
...
+{{- end }}
 metadata:
   name: {{ include "config-db.name" . }}
   labels: {{- include "config-db.labels" . | nindent 4 }}
 spec:
   replicas: {{ .Values.replicas }}
   selector:
     matchLabels: {{- include "config-db.selectorLabels" . | nindent 6 }}
+  {{- if and (eq .Values.db.external.enabled false) (eq .Values.db.embedded.persist true) }}
+  serviceName: {{ include "config-db.name" . }}
```
moshloop, line 15:
```suggestion
  serviceName: {{ include "config-db.name" . }}
  {{- if and (eq .Values.db.external.enabled false) (eq .Values.db.embedded.persist true) }}
```

chart/values.yaml
```diff
@@ -4,34 +4,43 @@
...
 
 configChangeRetentionDays: 60
 configAnalysisRetentionDays: 60
 
 # a list of configmaps to load scrape rules from, the configmap should have a single entry called "config.yaml"
 scrapeRuleConfigMaps:
   - config-db-rules
+
 db:
-  # Setting this to true will create a postgres stateful set for config-db to connect to.
```
moshloop, line 25: We seem to have lost the ability to deploy a postgres statefulset ?

adityathebe: I think you can still set `db.external.enabled: true` ?

adityathebe: need to add `create`

adityathebe:
@moshloop I'm not really sure why we get `operation not permitted` error when using embedded database.

- The Dockerfile correctly has the owner of `/opt/database` directory set to the new `catalog(1000)` user. I tested this on a docker container and the permission is correctly set.
- Inside Kubernetes, since we're mounting a volume to `/opt/database`, I have also set `fsGroup` to `1000` in the security context. Can also confirm that the process is run as `catalog` user.

Despite this, the error still occurs. 
Do you see anything obvious that I'm missing?

adityathebe:
@moshloop This is ready to go

PR #321, opened 2023-11-27
feat: kubernetes relationship by lookup

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/320


db/update.go
```diff
@@ -359,14 +361,19 @@ func relationshipResultHandler(relationships v1.RelationshipResults) error {
...
+		if relationship.RelatedConfigID != uuid.Nil {
+			relatedID = commonUtils.Ptr(relationship.RelatedConfigID.String())
+		} else {
+			relatedID, err = FindConfigItemID(relationship.RelatedExternalID)
+			if err != nil {
+				logger.Errorf("error fetching external config item(id=%s): %v", relationship.RelatedExternalID, err)
+				continue
+			}
+			if relatedID == nil {
+				logger.Warnf("related external config item(id=%s) not found.", relationship.RelatedExternalID)
```
moshloop, line 374:
```suggestion
				logger.Debugf("related external config item(id=%s) not found.", relationship.RelatedExternalID)
```

moshloop:
@adityathebe  Could we update the example fixture and maybe add an integration test?

db/config.go
```diff
@@ -87,6 +89,19 @@ func UpdateConfigItem(ci *models.ConfigItem) error {
 	return nil
 }
 
+// ConfigItemsIDs returns the uuid of config items which matches the given type, name & namespace
+func ConfigItemsIDs(ctx context.Context, configType, name, namespace string) ([]uuid.UUID, error) {
```
moshloop, line 93: Shouldn't this be FindConfigByNamespaceName?

scrapers/runscrapers_test.go
```diff
@@ -26,14 +30,85 @@ var _ = Describe("Scrapers test", func() {
...
+			scraperCtx := api.NewScrapeContext(gocontext.TODO(), gormDB, nil).WithScrapeConfig(&scrapeConfig)
+			_, err := RunScraper(scraperCtx)
+			Expect(err).To(BeNil())
+
+			var configRelationships []models.ConfigRelationship
+			err = gormDB.Find(&configRelationships).Error
+			Expect(err).To(BeNil())
+
+			Expect(len(configRelationships)).To(Equal(1))
+			Expect(configRelationships[0].Relation).To(Equal("ConfigMapConfigMap"))
```
moshloop, line 108: I think this should just be ConfigMap

adityathebe: Aren't we concatenating the kind of the two related objects? Eg (from azure demo):
```       
PodConfigMap                    
NodePod           
GrafanaConfigMap         
EKSNode                         
DaemonSetPod
```

moshloop: We shouldn't be, but we can tackle that later

adityathebe: https://github.com/flanksource/config-db/issues/323

PR #324, opened 2023-12-01
feat: change retention

### LINKED ISSUE #307: Change retention rules
drop all ReconciliationSucceeded changes older than 1d or when there are more than 10
```
kind:  Scraper:
retention: 
   changes:
       - name: ReconciliationSucceeded
          age: 1d
          count: 10

```

yashmehrotra:
Fixes #307


scrapers/run.go
```diff
@@ -20,14 +21,21 @@ func RunScraper(ctx api.ScrapeContext) (v1.ScrapeResults, error) {
...
+	if persistedID != nil {
+		// If error in any of the scrape results, don't delete old items
+		if len(results) > 0 && !v1.ScrapeResults(results).HasErr() {
 			if err := DeleteStaleConfigItems(ctx.DutyContext(), *persistedID); err != nil {
 				return nil, fmt.Errorf("error deleting stale config items: %w", err)
 			}
 		}
+
+		// Process change retention
+		for _, change := range ctx.ScrapeConfig().Spec.Retention.Changes {
```
moshloop, line 34: I think we should do this as a job, rather than at every scrape

Issue #319, opened 2023-11-26
Azure activity log


moshloop:
e.g. https://management.azure.com/subscriptions/0cd017bb-aa54-4611-b21f-ecf8daee0511/providers/microsoft.insights/eventtypes/management/values

PR #325, opened 2023-12-05
feat: azure activity logs

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/319


adityathebe:
Not sure why there are multiple events for the same action.

image

moshloop:
@adityathebe  what are we using as the change id ?

adityathebe:
image
---
image
---
image
---

scrapers/azure/azure.go
```diff
@@ -148,6 +180,117 @@ func (azure Scraper) Scrape(ctx api.ScrapeContext) v1.ScrapeResults {
...
+	clientFactory, err := armmonitor.NewClientFactory(azure.config.SubscriptionID, azure.cred, nil)
+	if err != nil {
+		return append(results, v1.ScrapeResult{Error: fmt.Errorf("failed to initiate arm monitor client: %w", err)})
+	}
+
+	// corelatedActivities keeps together events that belong to the same "uber operation"
+	// https://learn.microsoft.com/en-us/rest/api/monitor/activity-logs/list?view=rest-monitor-2015-04-01&tabs=HTTP#eventdata
+	var corelatedActivities = map[string][]activityChangeRecord{}
+
+	filter := fmt.Sprintf("eventTimestamp ge '%s'", time.Now().UTC().Add(activityLogTimespan).Format(time.RFC3339))
```
moshloop, line 211: I think we will need to maintain a maxTimestamp and only query after that time - like we do for Cloudtrail

scrapers/azure/azure.go
```diff
@@ -148,6 +180,117 @@ func (azure Scraper) Scrape(ctx api.ScrapeContext) v1.ScrapeResults {
 	return results
 }
 
+// azureActivityExcludeOperations is a list of activity logs we can avoid
+// as they do not mutate any resources.
+var azureActivityExcludeOperations = []string{
+	"Microsoft.ContainerService/managedClusters/listClusterAdminCredential/action",
```
moshloop, line 186: These should probably be exclusions at the crd level

adityathebe: hmm not sure. I only added these two in particular for debugging reasons but then kept it.
How would end users get the list of operations though? I couldn't find the list.

Still I have implemented this.

moshloop: It would normally by seeing lots of changes they are not interested in, and then adding them to the exclusions

PR #327, opened 2023-12-05
feat: establish azure resource relationship to resource group & subscription

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/318


adityathebe:
image

Issue #347, opened 2024-01-05
Kubernetes Linking


moshloop:
- [x] If  labels.   `eks.amazonaws.com/compute-type=ec2` && `spec:providerID: instance id://*/i-*` linked to EC2 node based on `i-*` number 
- [x] If there is a aws-auth cm, then insert its contents into the cluster JSON at aws-auth + extract the account id from the roles
- [x] If `labels.    alpha.eksctl.io/cluster-name: mission-control-demo-cluster` exists on any node, extract the cluster name and insert into cluster spec + link to EKS::Cluster
- [x] If there  aws-node pods from aws-node deaemson
    * [x]  extract AWS_ROLE_ARN env and link the node the pod is running on to that EKS::IAM:role
    * [x] extract VPC_ID and add it the Kubernetes::node, and Kubernetes::Cluster as labels and relationships 
    * [x] extract CLUSTER_NAME and add it Kubernetes::Cluster as spec, and link to Kubernetes::Cluster to AWS::EKS Cluster on cluster name and account id

PR #349, opened 2024-01-08
feat: Kubernetes Linking

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/347


scrapers/kubernetes/kubernetes.go
```diff
@@ -108,6 +113,119 @@ func (kubernetes KubernetesScraper) Scrape(ctx api.ScrapeContext) v1.ScrapeResul
...
+							if vpcID != "" {
+								labelsToAddToNode[nodeName] = map[string]string{
+									"vpc-id": vpcID,
+								}
+
+								if clusterScrapeResult, ok := results[0].Config.(map[string]any); ok {
+									clusterScrapeResult["vpc-id"] = vpcID
+								}
+
+								relationships = append(relationships, v1.RelationshipResult{
```
moshloop, line 162: I don't think we need to relate pod to cluster directly

Issue #356, opened 2024-01-22
Add Transform.exclude.filter


moshloop:
i.e. select which config types the fields should be dropped from

PR #362, opened 2024-01-23
feat: config field exclusion by types

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/356


moshloop:
@adityathebe can we add a test ?

moshloop:
@adityathebe I think class should be optional on the spec, and if empty default it to the type

adityathebe:
@moshloop Added tests and set default for config class

moshloop:
@adityathebe please rebase

Issue #360, opened 2024-01-22
AWS Resources linking


moshloop:
- [x] Instance parent to the VPC + subnet, region, zone
- [x] Subnet parent should be the VPC + links to Region, availability zone
- [x] VPC parent to account + region
- [x] EKS parent to VPC, + to subnets, instance roles, security groups, 
- [x] Everything without an explicit parent should have VPC or Account as parent

PR #368, opened 2024-01-25
feat: aws resource linking

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/360


scrapers/aws/aws.go
```diff
@@ -915,16 +1033,24 @@ func (aws Scraper) iamProfiles(ctx *AWSContext, config v1.AWS, results *v1.Scrap
 	tags := make(map[string]string)
 	tags["account"] = *ctx.Caller.Account
 	for _, profile := range profiles.InstanceProfiles {
+		// Use the first IAM role as the parent
```
moshloop, line 1036: I don't think we want to set the first iam role as parent

Issue #357, opened 2024-01-22
Add transform.relationships


moshloop:
Like Kubernetes relationships, but let a CEL expression return the Type, namespace, name tuples

```yaml
transform:
   relationships:
       - filter: type == 'Kubernetes::Application"
          expr:  [{name: 'a', namespace:'b', type: 'c']
```

adityathebe:
- (id/name) , type, agent, labels

PR #371, opened 2024-01-26
feat: customizable relationship formation

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/357


db/config.go
```diff
@@ -99,6 +99,32 @@ func UpdateConfigItem(ci *models.ConfigItem) error {
...
+	query := ctx.DB().Select("id, name, type")
+	if selector.Name != nil {
+		query = query.Where("name = ?", *selector.Name)
+	}
+
+	if selector.Type != nil {
+		query = query.Where("type = ?", *selector.Type)
+	}
+
+	if selector.Agent != nil {
```
moshloop, line 117: This should support `self` (i.e nil)  and an agent name

fixtures/kubernetes.yaml
```diff
@@ -6,6 +6,21 @@ spec:
...
     - clusterName: local-kind-cluster
       transform:
+        relationship:
+          # Link a service to a deployment (adjust the label selector accordingly)
+          - filter: config_type == "Kubernetes::Service"
+            type:
+              value: 'Kubernetes::Deployment'
+            name:
+              expr: |
+                has(config.JSON().spec.selector) && has(config.JSON().spec.selector.name) ? config.JSON().spec.selector.name : ''
```
moshloop, line 16: Don't we have content as a JSON object already ?

adityathebe: we don't. [This func](https://github.com/flanksource/config-db/blob/21f548970398b410462218deea9bc567289b909c/scrapers/kubernetes/kubernetes.go#L492) changes it to a string

api/v1/common.go
```diff
@@ -89,6 +92,139 @@ type TransformChange struct {
...
+	}
+
+	return false
+}
+
+// RelationshipSelectorCompiled is the evaluated output of RelationshipSelector.
+//
+// NOTE: The fields are pointers because we need to differentiate between
+// empty filter (null value) and empty result (output of the filter).
+type RelationshipSelectorCompiled struct {
```
moshloop, line 135: I would switch the name around, compiled is just RelationshipSelector, and the one with templating is RelationshipSelectorTemplate

adityathebe:
converting to draft for now. need to move the selector queries to duty.

PR #375, opened 2024-01-30
feat: make config.properties use filter

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/370


scrapers/processors/json.go
```diff
@@ -188,7 +188,26 @@ func (e Extract) Extract(inputs ...v1.ScrapeResult) ([]v1.ScrapeResult, error) {
...
+					input.Errorf("expected a boolean but property filter returned (%s)", response)
+					continue
+				} else if !boolVal {
+					continue
+				}
+			}
+
+			// Need to perform a deep clone otherwise this will affect the base scraper
+			// and hence all the inputs that come after.
+			configProperty, err = utils.CloneWithJSON(configProperty)
```
moshloop, line 206: Why not just. `clone := configProperty`

scrapers/processors/json.go
```diff
@@ -188,7 +188,26 @@ func (e Extract) Extract(inputs ...v1.ScrapeResult) ([]v1.ScrapeResult, error) {
 			}
 		}
 
-		if properties, ok := input.BaseScraper.Properties[input.Type]; ok {
+		for _, configProperty := range input.BaseScraper.Properties {
+			if configProperty.Filter != "" {
+				if response, err := gomplate.RunTemplate(input.AsMap(), gomplate.Template{Expression: configProperty.Filter}); err != nil {
```
moshloop, line 193: I think input.AsMap should only be called once for the entire extraction

Issue #393, opened 2024-02-08
a matched change -> action is not saved


adityathebe:
i.e. if for a pod killing event, we match that to delete action. The config is deleted but the actual "Killing" event isn't saved as a change.

Issue #392, opened 2024-02-07
Kubernetes resources being deleted incorrectly


moshloop:
https://incident-commander.demo.aws.flanksource.com/catalog/01884897-0ec9-18e6-c998-613bb109f880

PR #394, opened 2024-02-08
fix: save changes for events that match a change rule

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/393


adityathebe:
merging this to test this - https://github.com/flanksource/config-db/issues/392

PR #397, opened 2024-02-12
feat: move aws->kubernetes relationship to fixtures

adityathebe:

PR #398, opened 2024-02-13
feat: scrape azure scale set vms and link to k8s

yashmehrotra:
Azure VMs that run kubernetes nodes are part of a different API that we were never scraping before

### DIFF
--- a/scrapers/azure/azure.go
+++ b/scrapers/azure/azure.go
@@ -69,6 +69,10 @@ type Scraper struct {
 	ctx    context.Context
 	cred   *azidentity.ClientSecretCredential
 	config *v1.Azure
+
+	// Populate these first with list of all resource groups
+	// to fetch items grouped by resourceGroups
+	resourceGroups []string
 }
 
 func (azure Scraper) CanScrape(configs v1.ScraperSpec) bool {
@@ -127,6 +131,7 @@ func (azure Scraper) Scrape(ctx api.ScrapeContext) v1.ScrapeResults {
 		azure.config = &config
 		azure.cred = cred
 
+		// We fetch resource groups first as they are used to fetch further resources
 		results = append(results, azure.fetchResourceGroups()...)
 		results = append(results, azure.fetchVirtualMachines()...)
 		results = append(results, azure.fetchLoadBalancers()...)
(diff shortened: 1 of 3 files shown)


scrapers/azure/azure.go
```diff
@@ -521,11 +526,66 @@ func (azure Scraper) fetchVirtualMachines() v1.ScrapeResults {
...
+		}
+
+		for _, vmScaleSet := range nextPage.Value {
+			results = append(results, v1.ScrapeResult{
+				ConfigID:    vmScaleSet.Properties.UniqueID,
+				BaseScraper: azure.config.BaseScraper,
+				ID:          getARMID(vmScaleSet.ID),
+				Name:        deref(vmScaleSet.Name),
+				Config:      vmScaleSet,
+				ConfigClass: models.ConfigClassNode,
```
yashmehrotra, line 553: TODO: Should I add a ScaleSet Config Class ? Like AutoScaling group in AWS

scrapers/kubernetes/kubernetes.go
```diff
@@ -248,7 +248,14 @@ func extractResults(ctx context.Context, config v1.Kubernetes, objs []*unstructu
 							RelatedExternalID: v1.ExternalID{ExternalID: []string{ec2InstanceID}, ConfigType: "AWS::EC2::Instance"},
 							Relationship:      "EC2InstanceNode",
 						})
+					} else if strings.HasPrefix(providerID, "azure:///") && len(splits) > 0 {
+						relationships = append(relationships, v1.RelationshipResult{
```
adityathebe, line 252: I think VirtualMachine should be the config id and the node should be the related id since the relationship goes VirtualMachine -> Node

scrapers/kubernetes/kubernetes.go
```diff
@@ -248,7 +248,14 @@ func extractResults(ctx context.Context, config v1.Kubernetes, objs []*unstructu
 							RelatedExternalID: v1.ExternalID{ExternalID: []string{ec2InstanceID}, ConfigType: "AWS::EC2::Instance"},
 							Relationship:      "EC2InstanceNode",
 						})
+					} else if strings.HasPrefix(providerID, "azure:///") && len(splits) > 0 {
```
adityathebe, line 251: I think this relationship should be formed in the azure scraper instead.

adityathebe: https://github.com/flanksource/config-db/pull/397

yashmehrotra: In azure scraper, there is no way to know if a k8s node exists for that scaling set
I replicated the way we did for AWS, should we move the AWS ones as well ?

adityathebe: > In azure scraper, there is no way to know if a k8s node exists for that scaling set

I had a similar problem with IamRole -> kubernetes node linking. I added the iam role as a tag on the node and did a lookup for nodes with that tag from aws scraper.

PR #416, opened 2024-02-21
fix: add support for connectionPooler

dabasvibhor:


chart/values.yaml
```diff
@@ -24,6 +24,9 @@ scrapeRuleConfigMaps:
 
 db:
   runMigrations: true
+  connectionPooler:
+    enabled: false
+    secretKeyRef: ""
```
moshloop, line 29: Please see conments on canary-checkker pr

PR #418, opened 2024-02-21
feat: link aks cluster to kubernetes cluster & VM Scale set to the kubernetes nodes

adityathebe:
This PR adds necessary tags to the Azure resources and the related kubernetes resources so relationships can be formed between them.

related: https://github.com/flanksource/mission-control-registry/issues/51


api/v1/azure.go
```diff
@@ -20,7 +20,7 @@ type Azure struct {
 	Organisation   string           `yaml:"organisation" json:"organisation"`
 	ClientID       types.EnvVar     `yaml:"clientID,omitempty" json:"clientID,omitempty"`
 	ClientSecret   types.EnvVar     `yaml:"clientSecret,omitempty" json:"clientSecret,omitempty"`
-	TenantID       string           `yaml:"tenantID" json:"tenantID"`
+	TenantID       string           `yaml:"tenantID,omitempty" json:"tenantID,omitempty"`
```
adityathebe, line 23: Because this can come from the connection, it shouldn't be mandatory.

PR #421, opened 2024-02-22
feat: support change rules for diff changes as well

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/417


adityathebe:
image

PR #424, opened 2024-02-23
feat: save pipeline to save changes

adityathebe:
resolves: https://github.com/flanksource/mission-control-registry/issues/56

### DIFF
--- a/db/update.go
+++ b/db/update.go
@@ -27,6 +27,7 @@ import (
 	"github.com/lib/pq"
 	"github.com/patrickmn/go-cache"
 	"github.com/pkg/errors"
+	"github.com/samber/lo"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 )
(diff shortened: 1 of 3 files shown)


adityathebe:
Diff changes excluded & relationships were added

image

fixtures/kubernetes.yaml
```diff
@@ -26,6 +26,13 @@ spec:
                 filter(item, has(item.persistentVolumeClaim)).
                 map(item, {"type": "Kubernetes::PersistentVolumeClaim", "name": item.persistentVolumeClaim.claimName}).
                 toJSON()
+          # Link Argo Application to the resources
+          - filter: config_type == "kubernetes::Application" && config.apiVersion == "argoproj.io/v1alpha1"
+            expr: |
+              config.status.resources.map(item, {
+                "type": "Kubernetes::" + item.kind,
+                "name": item.name,                
+              }).toJSON()
```
adityathebe, line 35: @moshloop The relationship selector doesn't support `namespace` https://github.com/flanksource/config-db/blob/969c2cbea1d9625c73284c2ab70b0a781aae3dfb/api/v1/common.go#L173-L183

Maybe we need to add that.

db/update.go
```diff
@@ -145,28 +146,23 @@ func updateCI(ctx api.ScrapeContext, result v1.ScrapeResult) error {
...
 
 	return nil
 }
 
-func shouldExcludeChange(ctx dutyContext.Context, exclusions []string, changeResult v1.ChangeResult) (bool, error) {
+func shouldExcludeChange(ctx dutyContext.Context, result *v1.ScrapeResult, changeResult v1.ChangeResult) (bool, error) {
+	exclusions := result.BaseScraper.Transform.Change.Exclude
+
+	env := changeResult.AsMap()
+	env["config"] = result.Config
```
adityathebe, line 162: The actual scraped `config` is now passed to the exclusion filter cel-expression as it was needed to access the `apiVersion` in the argo exclusion filter.

fixtures/kubernetes.yaml
adityathebe: The `config_items` table does have `namespace` column but I think we're supposed to remove it https://github.com/flanksource/config-db/issues/294.

moshloop: Isn't it just a label ?

PR #427, opened 2024-02-27
fix: do not return error on should exclude change failing

yashmehrotra:
This error was occuring on aws instance:

```
2024-02-26T11:47:30.816	ERROR	failed to run scraper b8e78065-c9e4-40c8-bcf0-de3a39423092: failed to save results: failed to update db: [Endpoints/01884897-0e22-c8c9-82f3-628d3c5cf715] failed to save 1 changes: failed to evaluate change exclusion expression(config_type == "Kubernetes::Endpoints" && details.message == "metadata.annotations.endpoints.kubernetes.io/last-change-trigger-time"): error evaluating expression config_type == "Kubernetes::Endpoints" && details.message == "metadata.annotations.endpoints.kubernetes.io/last-change-trigger-time": no such key: message: no such key: message
```

As we return an error here, the entire scraping is marked as failed, none of the items are saved and in turn old items get deleted due to our scrape handling logic

Since this is a user input, we should just log and move on


adityathebe:
How would the user be aware of such errors in that case?

yashmehrotra:
> How would the user be aware of such errors in that case?

Right now logging, eventually Job history when we migrate to duty.Job

PR #429, opened 2024-02-28
fix: transform & deleting stale config

adityathebe:


scrapers/run.go
```diff
@@ -61,7 +61,11 @@ func UpdateStaleConfigItems(ctx api.ScrapeContext, results v1.ScrapeResults) err
 	// if the item was re-discovered in this run.
 	if val := ctx.Value(contextKeyScrapeStart); val != nil {
 		if start, ok := val.(time.Time); ok {
-			query := `UPDATE config_items SET deleted_at = NULL WHERE deleted_at IS NOT NULL AND ((NOW() - updated_at) <= INTERVAL '1 SECOND' * ?)`
+			query := `UPDATE config_items
```
adityathebe, line 64: added `AND deleted_at != updated_at` because any stale item that's deleted above is marked as recently updated and it's immeidately undeleted.

adityathebe: OR We could avoid updating the `updated_at` column when an item is deleted

api/v1/common.go
```diff
@@ -243,7 +243,7 @@ type RelationshipConfig struct {
 }
 
 type Transform struct {
-	Script Script `yaml:",inline" json:",inline"`
+	Script `yaml:",inline" json:",inline"`
```
adityathebe, line 246: This was breaking unmarshalling and wasn't accepting any scripts.

scrapers/run.go
moshloop: I think we should avoid updating the updated_at column where possible

adityathebe: https://github.com/flanksource/duty/pull/555

Issue #431, opened 2024-02-29
feat: stale config deletion using last_scraped time.


adityathebe:
- [ ] updated_at should only be updated when the actual update to the config happens
- [ ] the stale timeout should be per scraper and not global
- [ ] only save the last_scraped time if the scraper has stale timeout enabled.

PR #435, opened 2024-03-04
feat: stale config deletion using last_scraped time

yashmehrotra:
Fixes: https://github.com/flanksource/config-db/issues/431

Depends on: https://github.com/flanksource/duty/pull/567


api/v1/types.go
```diff
@@ -42,18 +42,20 @@ type RetentionSpec struct {
...
-	AzureDevops    []AzureDevops    `json:"azureDevops,omitempty" yaml:"azureDevops,omitempty"`
-	GithubActions  []GitHubActions  `json:"githubActions,omitempty" yaml:"githubActions,omitempty"`
-	Azure          []Azure          `json:"azure,omitempty" yaml:"azure,omitempty"`
-	SQL            []SQL            `json:"sql,omitempty" yaml:"sql,omitempty"`
-	Trivy          []Trivy          `json:"trivy,omitempty" yaml:"trivy,omitempty"`
-	Retention      RetentionSpec    `json:"retention,omitempty"`
+	LogLevel string `json:"logLevel,omitempty"`
+	Schedule string `json:"schedule,omitempty"`
+	// TODO: Better variable name, and it should be enabled by default and disable on user change
+	EnableStaleTimeout bool             `json:"enableStaleTimeout"`
```
yashmehrotra, line 48: @moshloop Can you suggest a better field name here ?

Also I think this should be the  default behaviour and stale item deletion should be set intentionally

scrapers/stale.go
```diff
@@ -23,8 +23,16 @@ func DeleteStaleConfigItems(ctx api.ScrapeContext, scraperID uuid.UUID) error {
 		}
 	}
 
-	if parsed, err := duration.ParseDuration(StaleTimeout); err != nil {
-		return fmt.Errorf("failed to parse stale timeout %s: %w", StaleTimeout, err)
+	staleTimeout := ctx.ScrapeConfig().Spec.Retention.StaleItemAge
+	if staleTimeout == "keep" {
+		ctx.DutyContext().Tracef("Skipping deletetion of stale config items")
```
moshloop, line 28:
```suggestion
```

scrapers/stale.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 var (
-	StaleTimeout string
+	DefaultStaleTimeout = "30m"
```
moshloop, line 15: Can we make this a property and change the default to 24h ?

PR #451, opened 2024-03-12
feat: support kubeconfig

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/445


scrapers/kubernetes/kubernetes.go
```diff
@@ -487,7 +488,11 @@ func updateOptions(ctx context.Context, opts *options.KetallOptions, config v1.K
 			return nil, err
 		}
 
-		opts.GenericCliFlags.KubeConfig = &val
+		if strings.HasPrefix(val, "/") {
```
moshloop, line 491: Should probably do an os.stat on it to check if it exists

moshloop: Except the user could specify a relative URL

PR #453, opened 2024-03-13
fix: add capability to specify additional volumeMounts

dabasvibhor:


moshloop:
@dabasvibhor - this should not be required, the latest version config-db supports getting the kubeconfig from configmap/secret

PR #460, opened 2024-03-17
Scrape improvements

moshloop:


api/cache.go
```diff
@@ -0,0 +1,110 @@
...
+		return item.ID, nil
+	}
+	return "", nil
+}
+
+func (t *TempCache) Find(typ, id string) (*models.ConfigItem, error) {
+	typ = strings.ToLower(typ)
+	id = strings.ToLower(id)
+
+	if strings.HasPrefix(typ, "kubernetes::") && uuid.Validate(id) == nil {
```
yashmehrotra, line 38: We have a const for this https://github.com/flanksource/config-db/blob/main/scrapers/kubernetes/kubernetes.go#L29

api/context.go
```diff
@@ -46,6 +53,8 @@ type ScrapeContext interface {
 type scrapeContext struct {
 	context.Context
 
+	temp *TempCache
```
yashmehrotra, line 56: tempCache or tCache

db/config.go
```diff
@@ -241,19 +185,10 @@ func GetJSON(ci models.ConfigItem) []byte {
...
-		if err != nil {
-			return err
-		}
-
-	}
-	return nil
+	return db.Clauses(clause.OnConflict{
+		Columns:   []clause.Column{{Name: "config_id"}, {Name: "related_id"}, {Name: "selector_id"}},
+		DoNothing: true,
+	}).CreateInBatches(relationships, 200).Error
```
yashmehrotra, line 191: Are we sure this works over the existing implementation ?

yashmehrotra: IIRC, in a transaction if multiple onconflict update statements  for the same constraint are present it throws an error

moshloop: There is no update, on conflict to nothing. 

There are cases where we need to handle the removal of the relationship, i.e. the direction changes but an update would help here anyway

PR #472, opened 2024-04-02
chore: correct namespace for scrape config from crd

yashmehrotra:


yashmehrotra:
@moshloop Merging this FYI

Issue #492, opened 2024-04-18
Support new is-healthy values


moshloop:
- add health and ready status

PR #494, opened 2024-04-22
feat: add health & ready to config items

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/492


scrapers/kubernetes/kubernetes.go
```diff
@@ -381,7 +382,7 @@ func extractResults(ctx context.Context, config v1.Kubernetes, objs []*unstructu
 		}
 
 		// Evicted Pods must be considered deleted
-		if obj.GetKind() == "Pod" && status == string(health.HealthStatusDegraded) {
+		if obj.GetKind() == "Pod" && string(resourceHealth.Status) == string(health.HealthStatusDegraded) {
```
moshloop, line 385: I don't think we should check status here at all

Issue #495, opened 2024-04-22
On startup, all scrapers are run twice


adityathebe:
One instance is run by the cron scheduler & the other instance is started by the k8s reconciler.

PR #502, opened 2024-04-23
fix: don't start scraper job from the reconciler on startup

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/495


db/config_scraper.go
```diff
@@ -68,15 +68,38 @@ func DeleteScrapeConfig(ctx api.ScrapeContext, id string) error {
 }
 
 func PersistScrapeConfigFromCRD(ctx context.Context, scrapeConfig *v1.ScrapeConfig) (bool, error) {
+	var changed bool
+
+	spec, err := utils.StructToJSON(scrapeConfig.Spec)
+	if err != nil {
+		return changed, err
```
yashmehrotra, line 75:
```suggestion
		return changed, fmt.Errorf("error converting to json: %w")
```

Issue #497, opened 2024-04-22
Add watches on specific resources (exclude them from incremental scrape of events)


moshloop:
Watch by default

- Pods
- Depoyments / Statefulsets, Daemset, Replicaset. Job, Cronjob

PR #509, opened 2024-04-25
feat: watch kubernetes resources

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/497


scrapers/kubernetes/events_watch.go
```diff
@@ -22,8 +26,63 @@ var (
...
+		var err error
+		ctx, err = applyKubeconfig(ctx, *config.Kubeconfig)
+		if err != nil {
+			return fmt.Errorf("failed to apply kube config")
+		}
+	}
+
+	var channels []<-chan watch.Event
+	for _, k := range config.WatchKinds {
+		// TODO: Refactor client creation
```
moshloop, line 51: This should move out of the loop

scrapers/kubernetes/events_watch.go
```diff
@@ -22,8 +26,63 @@ var (
 
 	// WatchEventBuffers stores a sync buffer per kubernetes config
 	WatchEventBuffers = make(map[string]chan v1.KubernetesEvent)
+
+	WatchResourceBufferSize = 5000
+
+	// WatchEventBuffers stores a sync buffer per kubernetes config
+	WatchResourceBuffer = make(map[string]chan *unstructured.Unstructured)
```
moshloop, line 33: Don't these need to be per scrapeconfig ?

adityathebe: it's that way already.

PR #511, opened 2024-04-26
fix: incremental scraper

adityathebe:
use the existing kube client by GroupVersionKind instead of ketall to fetch resources individually.
`ketall.KetOne` is unoptimized for high volume calls and introduces massive throttling.

### DIFF
--- a/scrapers/kubernetes/kubernetes.go
+++ b/scrapers/kubernetes/kubernetes.go
@@ -13,11 +13,15 @@ import (
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
 	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


scrapers/kubernetes/kubernetes.go
```diff
@@ -85,7 +106,7 @@ func (kubernetes KubernetesScraper) IncrementalScrape(ctx api.ScrapeContext, con
 		}
 	}
 
-	ctx.DutyContext().Logger.V(5).Infof("found %d objects for %d ids", len(objects), len(events))
+	ctx.DutyContext().Logger.V(1).Infof("found %d objects for %d events", len(objects)-len(events), len(events))
```
moshloop, line 109:
```suggestion
	ctx.DutyContext().Logger.V(3).Infof("found %d objects for %d events", len(objects)-len(events), len(events))
```

PR #513, opened 2024-04-27
fix: correct namespace for aws scraper

yashmehrotra:


yashmehrotra:
@moshloop Merging this

Also, we probably should have better (or more explicit namespace handling in ctx), right now it can come from multiple different places and hard to guess where it is coming from

PR #523, opened 2024-05-03
feat: change mapping action & summary

adityathebe:


scrapers/cron.go
```diff
@@ -142,6 +142,7 @@ func scheduleScraperJob(sc api.ScrapeContext) error {
 		Schedule:     schedule,
 		Singleton:    true,
 		JobHistory:   true,
+		RunNow:       true,
```
moshloop, line 145: We don't want to run this now, if config-db crashes then it will run all scrapers on startup even though some of them are only meant to run daily

api/v1/common.go
```diff
@@ -96,6 +96,12 @@ type ChangeMapping struct {
 	Filter string `json:"filter,omitempty"`
 	// Type is the type to be set on the change
 	Type string `json:"type,omitempty"`
+	// Action allows performing actions on the corresponding config item
+	// based on this change. Example: You can map EC2 instance's "TerminateInstances"
+	// change event to delete the corresponding EC2 instance config.
+	Action ChangeAction `json:"action,omitempty"`
```
moshloop, line 102: Can we add a list of allowed actions ?

PR #532, opened 2024-05-07
feat: check if migrations have run on startup

adityathebe:
resolves:  https://github.com/flanksource/config-db/issues/529

depends on: https://github.com/flanksource/duty/pull/749

### DIFF
--- a/cmd/operator.go
+++ b/cmd/operator.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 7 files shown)


cmd/operator.go
```diff
@@ -49,6 +51,12 @@ func run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to load properties: %v", err)
 	}
 
+	if ok, err := duty.HasMigrationRun(ctx, api.DefaultContext.Pool()); err != nil {
+		return fmt.Errorf("failed to check if migrations have run: %w", err)
+	} else if !ok {
+		return errors.New("migrations haven't run yet")
```
moshloop, line 57:
```suggestion
		return errors.New("Migrations not run, waiting for mission-control pod to start")
```

PR #549, opened 2024-05-16
Feat/changes rate limiter

adityathebe:


db/update.go
```diff
@@ -276,7 +276,11 @@ func extractChanges(ctx api.ScrapeContext, result *v1.ScrapeResult, ci *models.C
 		if changeResult.UpdateExisting {
 			updates = append(updates, change)
 		} else {
-			newOnes = append(newOnes, change)
+			if ok, err := ctx.TempCache().IsChangePersisted(change.ConfigID, change.ExternalChangeId); err != nil {
```
moshloop, line 279: This is going to be extremely slow - If we keep the on conflict do nothing, and add a returning config_id we can just increment the rate limit counter  after insertion

adityathebe: but the rate limit needs to happen before insertion. 

We must know which changes shouldn't account for rate limits.
Maybe a one query to filter out existing changes should do the job.

db/changes.go
```diff
@@ -10,3 +31,208 @@ func GetWorkflowRunCount(ctx api.ScrapeContext, workflowID string) (int64, error
...
+	rateLimitedConfigsPerScraper.Store(scraperID, rateLimitedConfigs)
+
+	return passingNewChanges, newlyRateLimited, nil
+}
+
+func syncCurrentlyRateLimitedConfigs(ctx api.ScrapeContext, window time.Duration) (map[string]struct{}, error) {
+	query := `WITH latest_changes AS (
+		SELECT
+			DISTINCT ON (config_id) config_id,
+			change_type
```
moshloop, line 114: The rate limit should be by config id, not config id and type. - We can just do`select config_id, count(*) from config_changes where scraper_id = ? and created_at > now() - window`

db/changes.go
```diff
@@ -10,3 +31,208 @@ func GetWorkflowRunCount(ctx api.ScrapeContext, workflowID string) (int64, error
...
+		configRateLimiters[configID] = rateLimiter
+	}
+
+	return rows.Err()
+}
+
+// filterOutPersistedChanges returns only those changes that weren't seen in the db.
+func filterOutPersistedChanges(ctx api.ScrapeContext, changes []*models.ConfigChange) ([]*models.ConfigChange, error) {
+	// use cache to filter out ones that we've already seen before
+	changes = lo.Filter(changes, func(c *models.ConfigChange, _ int) bool {
```
moshloop, line 192: This cache would be too big,  Why can't we just `insert on conflict do nothing returning config_id` - That will return  the config_id for changes that were inserted ?

db/changes.go
adityathebe: The rate limit is on config_id only.

https://github.com/flanksource/config-db/pull/549/files#diff-a359b758685582dec24c6f5672655ddff1c9392e2c5dfebe38eb874cc559c1e0R124

This query finds all the configs of the given scraper whose latest change was "TooManyChanges".

PR #583, opened 2024-05-23
chore: convert policy doc to json

yashmehrotra:
Fixes: https://github.com/flanksource/config-db/issues/126

### DIFF
--- a/scrapers/aws/aws.go
+++ b/scrapers/aws/aws.go
@@ -1074,6 +1074,13 @@ func (aws Scraper) iamProfiles(ctx *AWSContext, config v1.AWS, results *v1.Scrap
 			return
 		}
 
+		// We need to cast roles as []map[string]any to update the policy doc
+		var profileRoles []map[string]any
+		for _, r := range profileMap["Roles"].([]any) {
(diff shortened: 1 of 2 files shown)


scrapers/aws/aws.go
```diff
@@ -1066,12 +1068,42 @@ func (aws Scraper) iamProfiles(ctx *AWSContext, config v1.AWS, results *v1.Scrap
...
+
+		// We need to cast roles as []map[string]any to update the policy doc
+		var profileRoles []map[string]any
+		for _, r := range profileMap["Roles"].([]any) {
+			profileRoles = append(profileRoles, r.(map[string]any))
+		}
+		profileMap["Roles"] = profileRoles
+
+		for _, role := range profileMap["Roles"].([]map[string]any) {
+			if val, exists := role["AssumeRolePolicyDocument"]; exists {
```
yashmehrotra, line 1085: @moshloop @adityathebe Is there a better way to do this ?

`for _, role := range profileMap["Roles"].([]map[string]any) {`
will not work directly as it of type `[]any`

Issue #577, opened 2024-05-22
aws scraper graph


adityathebe:
```
{"level":"debug","ts":"2024-05-22T11:51:24.160044263Z","logger":"mission-control/aws","msg":"unable to get the path for config(id=37643963-3231-6337-3439-643434646634 name=AWS::AvailabilityZoneID type=euw1-az2 external_id=euw1-az2): target vertex not reachable from source"}
{"level":"debug","ts":"2024-05-22T11:51:24.161858216Z","logger":"mission-control/aws","msg":"unable to get the path for config(id=64333530-3933-3033-3865-343764316231 name=AWS::AvailabilityZoneID type=euw1-az3 external_id=euw1-az3): target vertex not reachable from source"}
{"level":"debug","ts":"2024-05-22T11:51:24.163961286Z","logger":"mission-control/aws","msg":"unable to get the path for config(id=38653165-3532-3461-3233-666131323963 name=AWS::AvailabilityZoneID type=euw1-az1 external_id=euw1-az1): target vertex not reachable from source"}
{"level":"debug","ts":"2024-05-22T11:51:24.173574712Z","logger":"mission-control/aws","msg":"unable to get the path for config(id=36386164-3930-3031-3232-386265306239 name=AWS::AvailabilityZoneID type=use1-az4 external_id=use1-az4): target vertex not reachable from source"}
{"level":"debug","ts":"2024-05-22T11:51:24.175290391Z","logger":"mission-control/aws","msg":"unable to get the path for config(id=30623730-3336-6662-6439-306162623034 name=AWS::AvailabilityZoneID type=use1-az6 external_id=use1-az6): target vertex not reachable from source"}
{"level":"debug","ts":"2024-05-22T11:51:24.176728414Z","logger":"mission-control/aws","msg":"unable to get the path for config(id=31323663-3961-3162-3465-666433386230 name=AWS::AvailabilityZoneID type=use1-az1 external_id=use1-az1): target vertex not reachable from source"}
{"level":"debug","ts":"2024-05-22T11:51:24.178213983Z","logger":"mission-control/aws","msg":"unable to get the path for config(id=36626131-3136-3765-6236-336364353735 name=AWS::AvailabilityZoneID type=use1-az2 external_id=use1-az2): target vertex not reachable from source"}
[...]

PR #596, opened 2024-05-29
fix: paths for availability zone & zone ID in aws scraper

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/577

We use a virtual root to deal with graphs with multiple parent-less items.

Fixes:
- [x] no need to manually define what config item is the root
- [x] can handle result set with multiple roots (example: results from aws scraper where the account config & all the configs for the aws regions are root)
- [x] parent of availability zone id

### DIFF
--- a/db/update.go
+++ b/db/update.go
@@ -253,9 +253,6 @@ func extractChanges(ctx api.ScrapeContext, result *v1.ScrapeResult, ci *models.C
 				return nil, nil, fmt.Errorf("failed to get config from change (externalID=%s): %w", change.GetExternalID(), err)
 			} else if ci != "" {
 				change.ConfigID = ci
-			} else if ci == "" {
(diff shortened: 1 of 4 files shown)


db/update.go
```diff
@@ -709,14 +706,21 @@ func extractConfigsAndChangesFromResults(ctx api.ScrapeContext, scrapeStartTime
...
 	for _, ci := range allConfigs {
 		if ci.ParentID != nil {
 			continue // existing item. Parent is already set.
 		}
 
 		if ci.ParentExternalID == "" || ci.ParentType == "" {
-			continue
+			if ci.ID == root || ctx.ScrapeConfig().IsCustom() {
+				continue
+			} else if lo.FromPtr(ci.Type) == v1.AWSRegion {
```
moshloop, line 718: Can't we handle this in the scraper itself? Maybe create a global root for AWS?

PR #615, opened 2024-06-05
feat: use shared informers to watch resource

adityathebe:


utils/kube/kube.go
```diff
@@ -88,13 +122,52 @@ func GetClientByGroupVersionKind(cfg *rest.Config, apiVersion, kind string) (dyn
 	return dc.Resource(mapping.Resource), nil
 }
 
+var clientSetCache = cache.New(time.Hour*24, time.Hour*24)
+
+func ClientSetFromRestConfig(restConfig *rest.Config) (*kubernetes.Clientset, error) {
```
moshloop, line 127: Shouldn't we be caching by kubeconfig and not rest config ?

scrapers/kubernetes/events_watch.go
```diff
@@ -37,8 +38,27 @@ var (
...
+
+	// The object returned by the informers do not have kind and apiversion set
+	m["kind"] = resource.Kind
+	m["apiVersion"] = resource.ApiVersion
+
+	return &unstructured.Unstructured{Object: m}, nil
+}
+
 // WatchResources watches Kubernetes resources
 func WatchResources(ctx api.ScrapeContext, config v1.Kubernetes) error {
```
moshloop, line 60: I think this should be externalised in WatchManager - if multiple kubernetes scrapers (with the same kubeconfig) are configured they should all use the same informers - Likewise updating a kubernetes scrapeconfig shouldn't make changes to the informers (unless it is to ask for them to be created

utils/kube/kube.go
```diff
@@ -38,6 +40,18 @@ import (
 	"k8s.io/client-go/util/homedir"
 )
 
+var kubeClientCreatedCount = prometheus.NewCounterVec(
```
moshloop, line 43: Should this be a ctx.Counter() ?

adityathebe: kube client creation is called in a bunch of places and they don't pass in the context. So I did it this way.

scrapers/kubernetes/events_watch.go
```diff
@@ -31,61 +31,59 @@ var (
...
 	WatchResourceBuffer = sync.Map{}
-
-	// DeleteResourceBuffer stores a buffer per kubernetes config
-	// that contains the ids of resources that have been deleted.
-	DeleteResourceBuffer = sync.Map{}
 )
 
-// WatchResources watches Kubernetes resources
+func getUnstructuredFromInformedObj(resource v1.KubernetesResourceToWatch, obj any) (*unstructured.Unstructured, error) {
+	b, err := json.Marshal(obj)
```
moshloop, line 37: Why do we need to marshal and unmarshall - It should be unstructured already ?

scrapers/kubernetes/informers.go
```diff
@@ -0,0 +1,206 @@
...
+
+	go func() {
+		informer.Run(stopper)
+		ctx.Logger.V(1).Infof("stopped shared informer for: %v", watchResource)
+	}()
+	return nil
+}
+
+// get returns an existing shared informer instance or creates & returns a new one.
+func (t *SharedInformerManager) get(ctx api.ScrapeContext, kubeconfig, apiVersion, kind string) (cache.SharedInformer, chan struct{}, bool) {
```
moshloop, line 101:
```suggestion
func (t *SharedInformerManager) getOrCreate(ctx api.ScrapeContext, kubeconfig, apiVersion, kind string) (cache.SharedInformer, chan struct{}, bool) {
```

scrapers/kubernetes/informers.go
```diff
@@ -0,0 +1,206 @@
...
+
+	cacheKey := apiVersion + kind
+
+	if val, ok := t.cache[kubeconfig]; ok {
+		if data, ok := val[cacheKey]; ok {
+			return data.informer, data.stopper, false
+		}
+	}
+
+	factory := informers.NewSharedInformerFactory(ctx.Kubernetes(), 0)
```
moshloop, line 113: Is this already prepopulated with kubeconfigs client ?

scrapers/kubernetes/events_watch.go
adityathebe: No, we receive concrete types like core `v1.Pod`

scrapers/kubernetes/informers.go
adityathebe: ya we set it here https://github.com/flanksource/config-db/blob/fd1e4480187e72e388b0120022290c06947434c6/scrapers/kubernetes/events_watch.go#L134

PR #624, opened 2024-06-11
chore: add ecr push action

yashmehrotra:


chart/values.yaml
```diff
@@ -18,7 +18,8 @@ nameOverride: ''
 disablePostgrest: false
 
 image:
-  repository: docker.io/flanksource/config-db
+  repository: docker.io/flanksource
```
moshloop, line 21: Can we switch to global.imageRegistry

yashmehrotra: Done

chart/templates/deployment.yaml
```diff
@@ -39,7 +39,7 @@ spec:
         fsGroup: 1000
       containers:
         - name: {{ include "config-db.name" . }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          image: "{{ tpl .Values.global.image.repository . }}/{{ .Values.image.name }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
yashmehrotra, line 42: global.imageRegistry

Issue #639, opened 2024-06-13
Custom annotations


moshloop:
- config-db.flanksource.com/tag/cluster: abc
- config-db.flanksource.com/ignore: *
- config-db.flanksource.com/ignore-change: *
- config-db.flanksource.com/ignore-change-severity: info

PR #644, opened 2024-06-14
feat: custom annotations

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/639

### DIFF
--- a/scrapers/kubernetes/kubernetes.go
+++ b/scrapers/kubernetes/kubernetes.go
@@ -12,8 +12,6 @@ import (
 	"github.com/flanksource/commons/collections"
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
-	"github.com/ohler55/ojg/jp"
-	"github.com/ohler55/ojg/oj"
 	"github.com/samber/lo"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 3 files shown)


scrapers/kubernetes/kubernetes.go
```diff
@@ -144,16 +161,79 @@ func (kubernetes KubernetesScraper) Scrape(ctx api.ScrapeContext) v1.ScrapeResul
...
+	} else if item != nil && item.Config != nil {
+		expr, err := jp.ParseString("metadata.annotations")
+		if err != nil {
+			return "", "", err
+		}
+
+		data, err := oj.ParseString(*item.Config)
+		if err != nil {
+			return "", "", err
+		}
```
adityathebe, line 217: @moshloop I think we should put all annotations as labels to avoid this parsing.

moshloop: That would create too many labels and clutter the dropdowns

scrapers/kubernetes/kubernetes.go
```diff
@@ -29,11 +31,23 @@ import (
...
+	AnnotationExcludeConfig = "config-db.flanksource.com/exclude"
+
+	// AnnotationExcludeChangeByType contains the list of change types to exclude
+	AnnotationExcludeChangeByType = "config-db.flanksource.com/exclude-changes"
+
+	// AnnotationExcludeChangeBySeverity contains the list of severity for the change types to exclude
+	AnnotationExcludeChangeBySeverity = "config-db.flanksource.com/exclude-change-severity"
+
+	// AnnotationCustomTags contains the list of tags to add to the scraped config
+	AnnotationCustomTags = "config-db.flanksource.com/tags"
```
moshloop, line 48: These should be labelled ignore - (even if we use the term exclude in the code). 
We might want to rename our existing fields to ignore in the crd eventually - but grammatically exclude is still ok in that scenario

scrapers/kubernetes/kubernetes.go
moshloop: We can add the 2 change annotations as labels though - Searching on configs that have ignored changes still makes sense

PR #646, opened 2024-06-16
chore: hardcode service account & use ecr registry

yashmehrotra:


chart/values.yaml
```diff
@@ -19,7 +19,7 @@ nameOverride: ''
 disablePostgrest: false
 
 image:
-  name: config-db
+  name: flanksource/config-db
```
moshloop, line 22:
```suggestion
  name: flanksource/config-db
```
```suggestion
  name: { tpl .Values.global.imagePrefix | default "flanksource" . }}/config-db
```

.github/workflows/release.yml
```diff
@@ -110,8 +99,9 @@ jobs:
           REPOSITORY: config-db
           IMAGE_TAG: "v${{ needs.semantic-release.outputs.release-version }}"
         run: |
-          docker build -t $REGISTRY/$REGISTRY_ALIAS/$REPOSITORY:$IMAGE_TAG -f build/Dockerfile . 
+          docker build -t $REGISTRY/$REGISTRY_ALIAS/$REPOSITORY:$IMAGE_TAG -t $REGISTRY/$REGISTRY_ALIAS/$REPOSITORY:latest -f build/Dockerfile .
```
moshloop, line 102: I don't think we should push latest - it will prvent us turning on immutable image tags

PR #653, opened 2024-06-18
feat: count orphaned changes as error

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/652


scrapers/aws/cloudtrail.go
```diff
@@ -133,8 +135,7 @@ func (aws Scraper) cloudtrail(ctx *AWSContext, config v1.AWS, results *v1.Scrape
 		start = lastEventTime.(time.Time)
 	}
 	err := lookupEvents(ctx, &cloudtrail.LookupEventsInput{
-		StartTime:  &start,
-		MaxResults: ptr.Int32(1000),
```
adityathebe, line 137: The max possible value is 50 (which is the default)

PR #657, opened 2024-06-19
chore: cleaner error count for orphaned

yashmehrotra:
Having a high error count without any description does not make sense

image

### DIFF
--- a/db/update.go
+++ b/db/update.go
@@ -119,7 +119,8 @@ func updateCI(ctx api.ScrapeContext, result v1.ScrapeResult, ci, existing *model
 		result.Changes = []v1.ChangeResult{*changeResult}
 		if newChanges, _, orphanedChanges, err := extractChanges(ctx, &result, ci); err != nil {
 			return false, nil, err
-		} else {
+		} else if orphanedChanges > 0 {
+			logger.Warnf("orphaned changes found for existing cofig")
 			ctx.JobHistory().AddErrorf("%d changes were orphaned for config item: [%s/%s]", orphanedChanges, existing.ID, *existing.Name)
 			changes = append(changes, newChanges...)
 		}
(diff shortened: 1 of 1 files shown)


db/update.go
```diff
@@ -119,7 +119,7 @@ func updateCI(ctx api.ScrapeContext, result v1.ScrapeResult, ci, existing *model
 		if newChanges, _, orphanedChanges, err := extractChanges(ctx, &result, ci); err != nil {
 			return false, nil, err
 		} else {
-			ctx.JobHistory().ErrorCount += orphanedChanges
+			ctx.JobHistory().AddErrorf("%d changes were orphaned for config item: [%s/%s]", orphanedChanges, existing.ID, *existing.Name)
```
yashmehrotra, line 122: Let me know if we want to put the orphaned count as error count, will change .AddError to pass that as an arg since .AddError automatically increments count by 1

moshloop, line 122: wouldn't existing always be nil here?

yashmehrotra: We have used `existing.ID` on L104 and `existing.Name` just below this, so I don't think these would be nil else we'd having a lot of ref errors already

adityathebe:
@moshloop I think we should get this merged in until we have the scrape summary

db/update.go
adityathebe: We're only dealing with diff changes here so we'll never have an orphaned change at this stage.

moshloop:
@adityathebe - We can't it could potentially have hundreds or thousands of error messages into the DB for each scrape.

adityathebe:
> @adityathebe - We can't it could potentially have hundreds or thousands of error messages into the DB for each scrape.

@moshloop I think we'll only have one error max per scrape for any orphaned changes error.

adityathebe:
tested. can confirm only one error gets logged.

PR #662, opened 2024-06-21
fix: ignore events whose involved object's UID is not a UUID

adityathebe:


scrapers/kubernetes/kubernetes.go
```diff
@@ -295,6 +297,18 @@ func ExtractResults(ctx api.ScrapeContext, config v1.Kubernetes, objs []*unstruc
 				continue
 			}
 
+			if _, err := uuid.Parse(string(event.InvolvedObject.UID)); err != nil {
+				ids, err := db.FindConfigIDsByNamespaceNameClass(ctx.DutyContext(), event.InvolvedObject.Namespace, event.InvolvedObject.Name, event.InvolvedObject.Kind)
```
moshloop, line 301: I don't think this will work as there might be multiple clusters

Issue #667, opened 2024-06-26
Handle whitespaces in ignore-changes annotation


dabasvibhor:
The whitespace after the `,` is causing issues in the scraper and its not able to "ignore" `ResourceUpdated`
`config-db.flanksource.com/ignore-changes: Operation*, ResourceUpdated`

PR #670, opened 2024-06-26
fix: whitespace in kubernetes exclusions

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/667


scrapers/kubernetes/kubernetes.go
```diff
@@ -811,3 +811,14 @@ func parseAzureURI(uri string) (string, string) {
 
 	return subscriptionID, vmScaleSetID
 }
+
+func splitTrimmed(input, cut string) []string {
```
moshloop, line 815: Shouldn't this go into MatchItems ?

adityathebe: I thought of that initially but then you would lose the ability to actually match space.

Issue #668, opened 2024-06-26
Make HelmRelease / Argo App a hard parent


moshloop:
hardcode for now, we can make it more generic later

PR #671, opened 2024-06-26
feat: Helm direct parent

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/668


db/update.go
```diff
@@ -730,10 +730,6 @@ func extractConfigsAndChangesFromResults(ctx api.ScrapeContext, scrapeStartTime
 
 func setConfigParents(ctx api.ScrapeContext, parentTypeToConfigMap map[configExternalKey]string, allConfigs []*models.ConfigItem) error {
 	for _, ci := range allConfigs {
-		if ci.ParentID != nil {
```
adityathebe, line 733: Removing this as it'll prevent updating the parents of existing item.

PR #686, opened 2024-07-02
feat(aws): scrape cloudformation stacks

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/506

depends on: https://github.com/flanksource/is-healthy/pull/62


adityathebe:
image

scrapers/aws/aws.go
```diff
@@ -183,6 +184,42 @@ func (aws Scraper) sqs(ctx *AWSContext, config v1.AWS, results *v1.ScrapeResults
...
+	ctx.Logger.V(2).Infof("scraping CloudFormation stacks")
+
+	client := cloudformation.NewFromConfig(*ctx.Session)
+	stacks, err := client.ListStacks(ctx, &cloudformation.ListStacksInput{})
+	if err != nil {
+		results.Errorf(err, "failed to list CloudFormation stacks")
+		return
+	}
+
+	for _, stack := range stacks.StackSummaries {
```
moshloop, line 201: Can we also link to the resources created by the stack ?

adityathebe: I think we can do that in the registry?

```yaml
 - filter: config_type == 'AWS::CloudFormation::Stack'
   expr: |
     [{"labels": {"aws:cloudformation:stack-id": config["StackId"]}}].toJSON()
```

moshloop: I actually think the parent of the AWS resource should be the Cloudformation stack - the same way we do helm / argo

adityathebe: yes this will set the cloudformation stack has the parent.

adityathebe: Tested


image

Issue #699, opened 2024-07-04
ECS task definition


adityathebe:
- set creation date
- map status

PR #701, opened 2024-07-08
feat(ecs task definition): fetch inactive & deleting definitions

adityathebe:
* map creation/deletion timestamps
* map health & status

resolves: https://github.com/flanksource/config-db/issues/699

### DIFF
--- a/scrapers/aws/aws.go
+++ b/scrapers/aws/aws.go
@@ -280,12 +280,12 @@ func (aws Scraper) ecsClusters(ctx *AWSContext, config v1.AWS, results *v1.Scrap
 			continue
 		}
 
-		labels := make(map[string]string)
-		for _, tag := range cluster.Clusters[0].Tags {
-			labels[*tag.Key] = *tag.Value
-		}
-
 		for _, clusterInfo := range cluster.Clusters {
+			labels := make(map[string]string)
+			for _, tag := range clusterInfo.Tags {
+				labels[*tag.Key] = *tag.Value
+			}
+
 			*results = append(*results, v1.ScrapeResult{
 				Type:        v1.AWSECSCluster,
 				Labels:      labels,
@@ -341,6 +341,11 @@ func (aws Scraper) ecsServices(ctx *AWSContext, config v1.AWS, client *ecs.Clien
 			Relationship:      "ECSTaskDefinitionECSService",
 		})
 
+		labels := make(map[string]string)
+		for _, tag := range service.Tags {
+			labels[*tag.Key] = *tag.Value
+		}
+
 		*results = append(*results, v1.ScrapeResult{
 			Type:                v1.AWSECSService,
 			ID:                  *service.ServiceArn,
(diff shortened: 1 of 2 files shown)


scrapers/aws/aws.go
```diff
@@ -360,44 +362,61 @@ func (aws Scraper) ecsTasks(ctx *AWSContext, config v1.AWS, results *v1.ScrapeRe
...
+	var tdStatus ecsTypes.TaskDefinitionStatus
+	for _, status := range tdStatus.Values() {
+		ctx.Logger.V(3).Infof("scraping %s ECS task definitions", status)
+
+		var configStatus = models.HealthUnknown
+		switch status {
+		case ecsTypes.TaskDefinitionStatusActive:
+			configStatus = models.HealthHealthy
+		case ecsTypes.TaskDefinitionStatusInactive:
+			configStatus = models.HealthUnhealthy
```
moshloop, line 374: Inactive is more akin to deleted then unhealthy

adityathebe: This var is incorrectly named `status`. Changed it to 'configHealth'

moshloop: I think we should rather use GetTasks and Describe Tasks -  I don't think we care about the task definitions so much as the running task. - The API returns stopped tasks for up to an 1h - so we will pick up short tasks as well,.

scrapers/aws/aws.go
```diff
@@ -346,62 +354,143 @@ func (aws Scraper) ecsServices(ctx *AWSContext, config v1.AWS, client *ecs.Clien
...
-			describeTaskDefinitionOutput, err := client.DescribeTaskDefinition(ctx, &ecs.DescribeTaskDefinitionInput{
-				TaskDefinition: &taskDefinitionArn,
-			})
-			if err != nil {
-				results.Errorf(err, "failed to describe ECS task definition %s", taskDefinitionArn)
-				continue
-			}
+		describeTasksOutput, err := client.DescribeTasks(ctx, &ecs.DescribeTasksInput{
+			Cluster: &clusterArn,
+			Tasks:   tasks.TaskArns,
```
moshloop, line 390: I think this might need to be buffered in groups of 100 as DescribeTasks takes a max of task arns

scrapers/aws/aws.go
```diff
@@ -346,62 +354,143 @@ func (aws Scraper) ecsServices(ctx *AWSContext, config v1.AWS, client *ecs.Clien
...
+	for _, status := range tdStatus.Values() {
+		ctx.Logger.V(3).Infof("scraping %s ECS task definitions", status)
+
+		configHealth := models.HealthUnknown
+		switch status {
+		case ecsTypes.TaskDefinitionStatusActive:
+			configHealth = models.HealthHealthy
+		case ecsTypes.TaskDefinitionStatusInactive:
+			configHealth = models.HealthUnhealthy
+		}
```
moshloop, line 445: A task definition is not running - therefor it it doesn't have health - Only the running task does

adityathebe: makes sense. I initially added it because when only the `status` is set with an empty health, nothing was shown on the UI.

PR #705, opened 2024-07-08
feat: link kubernetes LB services to AWS ELB

adityathebe:
~~## fix source vertex not found error by namespacing all look ups to the scraper.~~

~~The source vertex not found error would occur only in cases where multiple clusters were being scraped. The lookups needed to be scoped to the given scraper.~~

~~_Example_: When forming a `Namespace default` -> `Deployment Nginx` relationship for **cluster A,** the lookup for external id `Kubernetes/Namespace//default` would return the ID of the `default namespace` from **cluster B.** This would cause failure when forming the graph as the edge couldn't be inserted since there's no source vertex in the graph.~~

~~resolves: https://github.com/flanksource/config-db/issues/703~~

### DIFF
--- a/scrapers/aws/aws.go
+++ b/scrapers/aws/aws.go
@@ -1264,7 +1264,7 @@ func (aws Scraper) loadBalancers(ctx *AWSContext, config v1.AWS, results *v1.Scr
 			Name:                *lb.LoadBalancerName,
 			Labels:              labels,
 			Tags:                tags,
-			Aliases:             []string{"AWSELB/" + arn, arn},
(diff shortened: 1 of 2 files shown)


adityathebe:
> I think we just need to change https://github.com/flanksource/config-db/blob/main/scrapers/kubernetes/kubernetes.go#L438 to use the resourceIdMap

The problem is that on incremental scrapes, the resource id map is nil. Or at least, we cannot guarantee that the namespace or possibly other parents as well are going to be in the resource id map.

I'll create a separate issue for this.

Issue #728, opened 2024-07-11
aws: region, availability zone and zone id shouldn't belong to a scraper.


adityathebe:
- If an account has opted in to a region then we form a relationship.

PR #741, opened 2024-07-11
feat(aws): region, zone & zone id are not account specific

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/728

image

image


adityathebe:
@moshloop we might have to add the depth slider in the graph. 

Selecting "soft" & "outgoing" on the region brings in everything whereas I just want to see which accounts are in the region.

PR #751, opened 2024-07-16
fix: kubernetes diff exclusion via annotation

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/726

### DIFF
--- a/api/v1/kubernetes.go
+++ b/api/v1/kubernetes.go
@@ -14,6 +14,21 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
+// ReservedAnnotations
+const (
+	// AnnotationIgnoreConfig excludes the object from being scraped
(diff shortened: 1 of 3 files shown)


api/v1/kubernetes.go
```diff
@@ -14,6 +14,21 @@ import (
...
+	// AnnotationIgnoreChangeByType contains the list of change types to ignore
+	AnnotationIgnoreChangeByType = "config-db.flanksource.com/ignore-changes"
+
+	// AnnotationIgnoreChangeBySeverity contains the list of severity for the change types to ignore
+	AnnotationIgnoreChangeBySeverity = "config-db.flanksource.com/ignore-change-severity"
+
+	// AnnotationCustomTags contains the list of tags to add to the scraped config
+	AnnotationCustomTags = "config-db.flanksource.com/tags"
+)
+
```
adityathebe, line 31: moved here due to import cycle between `config-db/db` and `scrapers/kubernetes`

Issue #676, opened 2024-06-28
Argo app hard parent 


adityathebe:
related: https://github.com/flanksource/config-db/issues/668

PR #753, opened 2024-07-17
feat: set argo resources as children

yashmehrotra:
Fixes: #676

### DIFF
--- a/db/update.go
+++ b/db/update.go
@@ -749,12 +749,15 @@ func extractConfigsAndChangesFromResults(ctx api.ScrapeContext, scrapeStartTime 
 		}
 	}
 
-	// Calculate the parents only after we have all the config items.
+	// Calculate the parents and children only after we have all the config items.
 	// This is because, on the first run, we don't have any configs at all in the DB.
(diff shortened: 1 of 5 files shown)


db/update.go
```diff
@@ -794,6 +797,35 @@ func setConfigParents(ctx api.ScrapeContext, parentTypeToConfigMap map[configExt
...
+			if child.ExternalID == "" || child.Type == "" {
+				continue
+			}
+
+			found, err := ctx.TempCache().Find(ctx, v1.ExternalID{ConfigType: child.Type, ExternalID: []string{child.ExternalID}})
+			if err != nil {
+				return err
+			}
+			if found == nil {
+				ctx.Logger.V(0).Infof("child:[%s/%s] not found for config [%s]", child.Type, child.ExternalID, ci)
```
moshloop, line 817:
```suggestion
if ctx.PropertyOn(false, "log.missing") {
				ctx.Logger.Tracef("child:[%s/%s] not found for config [%s]", child.Type, child.ExternalID, ci)
				}
```

Issue #769, opened 2024-07-19
Slack - Changes from messages


moshloop:
```yaml
slack:
  token: xxx
  channels:
    - alerts-*
  rules:
    - mapping:
        severity:
          expr: >
            ":large_red_circle:" in text ? "high" : ":large_orange_circle:" in text ? "med" : "info"
        type:
          expr: >
            regexp.Find(":\w+: \w: (.*)\n|:", text)
        created_at:
        summary:
          expr: text
      config:
        # sequential list of selectors to attach the change to
        - name:
            value: regexp.Find("http://(\w+)", text)
          type:
            value: MongoDB
          tags:
            a: b
        - type:
            value: AWS::Account
      filter:
        #  matchitems: * for any bot, or bot name
        bot: mongodb-atlas-alert-app
        user:
          name: '!*' # skip user message
          displayName:
        # CEL filter
        expr:
```
The mapping and filtering rules should be externalized so they can be used for generic HTTP / S3 / etc to change lookups

PR #772, opened 2024-07-19
feat(slack scraper): create types

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/769


api/v1/common.go
```diff
@@ -354,3 +359,26 @@ type Template struct {
...
+type ChangeExtractionVar struct {
+	Expr CelExpression `yaml:"expr,omitempty" json:"expr,omitempty"`
+
+	// Value is a static value
+	Value string `yaml:"value,omitempty" json:"value,omitempty"`
+}
+
+// ChangeExtractionConfigSelector is used to select a config item.
+// At least one of the fields must be specified.
+type ChangeExtractionConfigSelector struct {
```
moshloop, line 380: Shouldn't this be a EnvVarResourceSelector

fixtures/slack.yaml
```diff
@@ -0,0 +1,37 @@
...
+            name: slack-mission-control-bot
+            key: token
+      rules:
+        - regexp: name:\s*(?<name>.*?)\s+type:\s*(?<change_type>.*?)\s+is\s+.*\nseverity:\s*(?<severity>\w+)
+          filter:
+            bot: 'Notifier'
+            expr: text.contains('healthy') # match healthy/unhealthy
+          config:
+            - name:
+                expr: env.?name.orValue('')
```
adityathebe, line 25: @moshloop do you think we could add a `strict_regexp: bool` field in `rule` so we don't need to have this kind of condition null check?

i.e. all the capture fields need to find it's text.

api/v1/common.go
```diff
@@ -354,3 +371,58 @@ type Template struct {
...
+	if t.Value != "" {
+		return t.Value, nil
+	}
+
+	return gomplate.RunTemplate(env, gomplate.Template{Expression: string(t.Expr)})
+}
+
+// EnvVarResourceSelector is used to select a config item.
+// At least one of the fields must be specified.
+type EnvVarResourceSelector struct {
```
moshloop, line 407: This limits the options for config selection - can't we create a new  EnvVarResourceSelector in duty, that has a `EnvVarResourceSelector.Hydrate(ctx)` so that is reusable and you can attach changes to configs with any of the selector fields

fixtures/slack.yaml
moshloop: I think this should be a default but only for mapping - If the regex is supplied by default it uses that field, explicit mapping is in addition ontop of regex.

moshloop: for config lookup - I think that needs to be explicit

scrapers/changes/extraction.go
```diff
@@ -0,0 +1,135 @@
...
+	parsed, err := regexp.Compile(r)
+	if err != nil {
+		return nil, err
+	}
+	regexpCache.Store(r, parsed)
+
+	return parsed, nil
+}
+
+func MapChanges(ctx api.ScrapeContext, rule v1.ChangeExtractionRule, text string) ([]v1.ChangeResult, error) {
```
moshloop, line 34: Can we add unit tests for this ?

fixtures/slack.yaml
adityathebe: > If the regex is supplied by default it uses that field

You could have a scenario where the regexp is partially matched.

scrapers/slack/slack.go
```diff
@@ -0,0 +1,169 @@
...
+
+	opt := &GetConversationHistoryParameters{}
+	if config.Since != "" {
+		parsed, err := duration.ParseDuration(config.Since)
+		if err != nil {
+			results = append(results, v1.ScrapeResult{Error: fmt.Errorf("bad duration string %s: %w", config.Since, err)})
+			return results
+		}
+
+		opt.Oldest = time.Now().Add(-time.Duration(parsed)).Unix()
```
moshloop, line 68: I think this might need to be cached with the max timestamp during scrape

scrapers/slack/slack.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	return results
+}
+
+func (s Scraper) scrapeChannel(ctx api.ScrapeContext, config v1.Slack, client *SlackAPI, channel ChannelDetail) []v1.ScrapeResult {
+	var results v1.ScrapeResults
+
+	opt := &GetConversationHistoryParameters{}
+	if config.Since != "" {
```
moshloop, line 64: I think this should default to 7d if empty

Issue #44, opened 2022-05-23
HTTP Scraper


moshloop:
Create a new scraper, similar to the file scraper (which uses go-getter) except the http scraper uses HTTPConnection / client with support for

* Authentication / Oauth / Tls
* Headers
* Method
* Body

https://github.com/flanksource/canary-checker/blob/master/checks/http.go can. be used as a base

```yaml
http
  - type: $.response.type # $.response is body of the http response
    id: $.headers.id # headers is a http headers from the response
    url: https://domain/requires_auth
    method: POST
    body: |+
         {"hello": "world"}
    authentication:
         username:
            value: dummy
        password:
            secretMapRef:
                name: auth
                key: password

mrinalwahal:
Hey, @moshloop!

Can you please provide a description of what this issue is about? So that we can contribute and solve this.

The same goes for most of the issues labelled "good-first-issue."

Thanks! :)

moshloop:
@mrinalwahal  see updated

PR #783, opened 2024-07-26
feat: http scraper

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/44


scrapers/http/http.go
```diff
@@ -0,0 +1,134 @@
...
+		} else {
+			results = append(results, *result)
+		}
+	}
+
+	return results
+}
+
+func scrape(ctx api.ScrapeContext, spec v1.HTTP) (*v1.ScrapeResult, error) {
+	conn, err := spec.HTTPConnection.Hydrate(ctx, ctx.Namespace())
```
moshloop, line 37: Can we use https://github.com/flanksource/duty/blob/main/connection/http.go#L115 instead ?

PR #797, opened 2024-08-02
fix: handle missing config ids when forming relationships

yashmehrotra:


adityathebe:
@yashmehrotra this to handle cases where the parent is excluded from being scraped?

yashmehrotra:
@adityathebe Yeah, there could be cases where object is excluded from the scrape or not present (saw one case of deleted node but non deleted CNINode which uses the node as a owner reference)

adityathebe:
@yashmehrotra I think we probably should add "on fk constraint error do nothing" instead

https://github.com/flanksource/config-db/blob/68d84ede655d1d7135345cee3402e94c8dfb484a/db/config.go#L204-L209

db/update.go
```diff
@@ -605,7 +605,16 @@ func relationshipResultHandler(ctx api.ScrapeContext, relationships v1.Relations
 
 		var configID string
 		if relationship.ConfigID != "" {
-			configID = relationship.ConfigID
+			configItem, err := ctx.TempCache().Get(ctx, relationship.ConfigID)
+			if err != nil {
+				ctx.Errorf("error fetching external config item(id=%s): %v", relationship.ConfigID, err)
+				continue
+			}
+			if configItem == nil {
```
moshloop, line 613:
```suggestion
			if configItem == nil && logMissing {
```

db/update.go
```diff
@@ -620,7 +629,16 @@ func relationshipResultHandler(ctx api.ScrapeContext, relationships v1.Relations
 
 		var relatedID string
 		if relationship.RelatedConfigID != "" {
-			relatedID = relationship.RelatedConfigID
+			relatedCI, err := ctx.TempCache().Get(ctx, relationship.RelatedConfigID)
+			if err != nil {
+				ctx.Errorf("error fetching related config item(id=%s): %v", relationship.RelatedConfigID, err)
+				continue
+			}
+			if relatedCI == nil {
```
moshloop, line 637:
```suggestion
			if relatedCI == nil && logMissing {
```

PR #799, opened 2024-08-02
feat: run relationships selector and properties after transformation

adityathebe:
* set health, status & readiness if empty

### DIFF
--- a/api/v1/interface.go
+++ b/api/v1/interface.go
@@ -582,6 +582,24 @@ type ScrapeResult struct {
 	RelationshipSelectors []duty.RelationshipSelector `json:"-"`
 }
 
+// SetHealthIfEmpty sets the health, status & readiness of the scrape result
+// based on the config type.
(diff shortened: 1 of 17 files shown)


api/v1/interface.go
```diff
@@ -582,6 +582,24 @@ type ScrapeResult struct {
 	RelationshipSelectors []duty.RelationshipSelector `json:"-"`
 }
 
+// SetHealthIfEmpty sets the health, status & readiness of the scrape result
+// based on the config type.
+func (s ScrapeResult) SetHealthIfEmpty() ScrapeResult {
+	if s.Status == "" && s.Health == "" {
+		if strings.HasPrefix(s.Type, "Mongo::") {
+			s = s.WithHealthStatus(health.GetMongoHealth(s.ConfigMap()))
```
moshloop, line 590: This doesn't feel right,  I think we need a `health.GetHealthByKind('Mongo::Cluster', ...)` which should be called in the scraper

PR #802, opened 2024-08-05
feat: different retention period of agent's configs

adityathebe:
merge after: https://github.com/flanksource/duty/pull/935


adityathebe:
it's the same retention period for both.

PR #813, opened 2024-08-07
feat: dedup changes

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/16


db/update.go
```diff
@@ -356,6 +363,40 @@ func SaveResults(ctx api.ScrapeContext, results []v1.ScrapeResult) (v1.ScrapeSum
 	return saveResults(ctx, results)
 }
 
+func dedupChanges(ctx api.ScrapeContext, window time.Duration, changes []*models.ConfigChange) ([]*models.ConfigChange, []string, error) {
+	if len(changes) == 0 {
```
moshloop, line 367: I think we can just maintain cache of config.id + fingerprint -> change.id and then for each change get the existing change and update it + reset the cache expiry for that change

moshloop:
@adityathebe - Please can we implement the the fingerprint cache instead of DB lookups

db/update.go
```diff
@@ -420,7 +427,11 @@ func saveResults(ctx api.ScrapeContext, results []v1.ScrapeResult) (v1.ScrapeSum
 		return summary, fmt.Errorf("failed to update last scraped time: %w", err)
 	}
 
-	if err := ctx.DB().CreateInBatches(newChanges, configItemsBulkInsertSize).Error; err != nil {
+	dedupWindow := ctx.Properties().Duration("changes.dedup.window", time.Minute*30)
```
moshloop, line 430:
```suggestion
	dedupWindow := ctx.Properties().Duration("changes.dedup.window", time.Minute*60)
```

changes/fingerprint.go
```diff
@@ -64,7 +80,7 @@ func Fingerprint(change *models.ConfigChange) string {
 	hash := Hash(out)
 	logger.Infof("\n%s\n--->\n%s\n===>  %s", print(flat), print(out), hash)
```
moshloop, line 81:
```suggestion
```

moshloop (changes requested):

moshloop (approved):

PR #843, opened 2024-08-20
feat(aws connection): use from duty and support default regions

adityathebe:
merge after: https://github.com/flanksource/duty/pull/966


chart/crds/configs.flanksource.com_scrapeconfigs.yaml
```diff
@@ -240,7 +240,7 @@ spec:
                     region:
                       items:
                         type: string
-                      type: array
+                      type: string
```
moshloop, line 243: This is a breaking change and would require a new CRD version and migrator

adityathebe: a `Region []string` was added to the scraper for backward compatibility.
https://github.com/flanksource/config-db/pull/843/files#diff-0686d136b2b5aa74ce26452f9d9a4acacd0c5d0308865aa8e44b4c6837312980R17

adityathebe: ah apparently there was an error on schema gen

```
github.com/flanksource/config-db/api/v1:-: conflicting types in allOf branches in schema: string vs array
```

api/v1/common.go
```diff
@@ -302,28 +303,30 @@ func (auth Authentication) GetDomain() string {
...
 	SecretKey      types.EnvVar `yaml:"secretKey,omitempty" json:"secretKey,omitempty"`
-	Region         []string     `yaml:"region,omitempty" json:"region"`
-	Endpoint       string       `yaml:"endpoint,omitempty" json:"endpoint,omitempty"`
-	SkipTLSVerify  bool         `yaml:"skipTLSVerify,omitempty" json:"skipTLSVerify,omitempty"`
 	AssumeRole     string       `yaml:"assumeRole,omitempty" json:"assumeRole,omitempty"`
+	Endpoint       string       `yaml:"endpoint,omitempty" json:"endpoint,omitempty"`
+	// Skip TLS verify when connecting to aws
+	SkipTLSVerify bool `yaml:"skipTLSVerify,omitempty" json:"skipTLSVerify,omitempty"`
+
+	Regions []string `yaml:"region" json:"region"`
```
moshloop, line 318:
```suggestion
	Regions []string `yaml:"region,omitempty" json:"region, omitempty"`
```

moshloop: The region should be optional - If not specified use below to get the current region (but keep the function in duty)

```

func getAWSRegion() (string, error) {
	// Check if AWS_DEFAULT_REGION is set
	if region := os.Getenv("AWS_DEFAULT_REGION"); region != "" {
		return region, nil
	}

	// Check if AWS_EC2_METADATA_DISABLED is set to true
	if os.Getenv("AWS_EC2_METADATA_DISABLED") == "true" {
		return "", fmt.Errorf("EC2 metadata is disabled")
	}

	// Define the IMDSv2 token URL and region URL
	tokenURL := "http://169.254.169.254/latest/api/token"
	regionURL := "http://169.254.169.254/latest/meta-data/placement/region"

	// Create a new HTTP client with a timeout
	client := &http.Client{
		Timeout: 2 * time.Second,
	}

	// Try to get the region using IMDSv2
	req, err := http.NewRequest("PUT", tokenURL, nil)
	if err != nil {
		return "", err
	}
	req.Header.Set("X-aws-ec2-metadata-token-ttl-seconds", "21600")

	resp, err := client.Do(req)
	if err == nil && resp.StatusCode == http.StatusOK {
		defer resp.Body.Close()
		token := resp.Header.Get("X-aws-ec2-metadata-token")

		req, err = http.NewRequest("GET", regionURL, nil)
		if err != nil {
			return "", err
		}
		req.Header.Set("X-aws-ec2-metadata-token", token)

		resp, err = client.Do(req)
		if err == nil && resp.StatusCode == http.StatusOK {
			defer resp.Body.Close()
			var region string
			_, err := fmt.Fscanf(resp.Body, "%s", ®ion)
			if err != nil {
				return "", err
			}
			return region, nil
		}
	}

	// Fallback to IMDSv1 if IMDSv2 fails
	resp, err = client.Get(regionURL)
	if err != nil {
		return "", err
	}
	defer resp.Body.Close()

	if resp.StatusCode != http.StatusOK {
		return "", fmt.Errorf("failed to get region from IMDSv1")
	}

	var region string
	_, err = fmt.Fscanf(resp.Body, "%s", ®ion)
	if err != nil {
		return "", err
	}

	return region, nil
}
```

hack/generate-schemas/go.mod
```diff
@@ -0,0 +1,243 @@
+module github.com/flanksource/config-db/hack/generate-schemas
```
moshloop, line 1: These should be gitignored - they are generated on-demand by Make gen-schemas

Issue #836, opened 2024-08-16
CRD Sync


moshloop:
When crdSync is true, and the type is MissionControl:Playbook,Canary,Connection,Notification create and update the corresponding object in the db. 

This will enable things like generating a playbook for helm charts in artifacthub, or for terraform modules. 

```yaml
file:
   - type: MissionControl::Playbook
      url: github.com/flanksource/canary-checker
      paths:
         - **/mission-control-playbook.yaml
crdSync: true
```

PR #868, opened 2024-08-23
feat: crd sync

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/836


db/update.go
```diff
@@ -376,6 +377,143 @@ func SaveResults(ctx api.ScrapeContext, results []v1.ScrapeResult) (v1.ScrapeSum
...
+		if err != nil {
+			return err
+		}
+
+		switch strings.TrimPrefix(*config.Type, api.MissionControlConfigTypePrefix) {
+		case "ScrapeConfig":
+			scrapeConfig := dutyModels.ConfigScraper{
+				Name:   fmt.Sprintf("%s/%s", namespace, *config.Name),
+				Spec:   string(spec),
+				Source: dutyModels.SourceUI,
```
moshloop, line 412: This should be CRDSync

fixtures/data/echo-playbook.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
moshloop, line 1: Can we add a test for this ?

PR #869, opened 2024-08-26
fix: add namespace scope to fix kubeconfig for watch resources

yashmehrotra:
Also fixed duty bump errors


yashmehrotra:
@moshloop Self merging this

PR #877, opened 2024-08-29
feat: use rust via FFI for memory efficient diff generation

yashmehrotra:


db/diff.go
```diff
@@ -90,16 +99,32 @@ func generateDiff(newConf, prevConfig string) (string, error) {
...
 
-	// Compare again. They might be equal after normalization.
-	if newConf == prevConfig {
+	if before == after {
 		return "", nil
 	}
 
-	edits := myers.ComputeEdits("", before, after)
-	if len(edits) == 0 {
+	if isSet := os.Getenv("DISABLE_RUST_DIFFGEN"); isSet != "" {
```
moshloop, line 106:
```suggestion
	if properties.On(true, "diff.rust-gen") {
```

db/diff.go
```diff
@@ -1,8 +1,17 @@
 package db
```
moshloop, line 1: Can we put this into another file with a build tag, so that we can build without rust as well ?

yashmehrotra:
Closing in favour of #879

PR #879, opened 2024-08-29
 feat: use rust via FFI for memory efficient diff generation

yashmehrotra:


db/diff.go
```diff
@@ -90,16 +96,26 @@ func generateDiff(newConf, prevConfig string) (string, error) {
 		return "", fmt.Errorf("failed to normalize json for new config: %w", err)
 	}
 
-	// Compare again. They might be equal after normalization.
-	if newConf == prevConfig {
```
yashmehrotra, line 94: We compare them previously and normalizer func does not modify the strings

Issue #880, opened 2024-08-30
Intermittent duplicate key value 


moshloop:
2024-08-29T23:54:45 ERROR: config-db-6956d7d548-k57z2 config-db  ERROR >=[1ms] [rows:0] ERROR: duplicate key value violates unique constraint "config_changes_config_id_external_change_id_key" (SQLSTATE 23505) UPDATE "config_changes" SET "change_type"=$1$,"count"=count + $2$,"created_at"=NOW(),"created_by"=$3$,"details"=$4$,"diff"=$5$,"external_change_id"=$6$,"external_created_by"=$7$,"severity"=$8$,"source"=$9$,"summary"=$10$ WHERE id = $11$
2024-08-29T23:54:45 ERROR: config-db-6956d7d548-k57z2 config-db  Scraper{b8e78065-c9e4-40c8-bcf0-de3a39423092} failed to save results: failed to create deduped config changes: ERROR: duplicate key value violates unique constraint "config_changes_config_id_external_change_id_key" (SQLSTATE 23505): detail: Key (config_id, external_change_id)=(962f999c-a9bd-40a4-80bf-47c84b1ad750, 8e4260f3-0d56-4ea2-8510-5781642e6ef9) already exists.
2024-08-29T23:54:45 ERROR: config-db-6956d7d548-k57z2 config-db  Scraper{b8e78065-c9e4-40c8-bcf0-de3a39423092} error running scraper[kubernetes]: failed to save results: failed to create deduped config changes: ERROR: duplicate key value violates unique constraint "config_changes_config_id_external_change_id_key" (SQLSTATE 23505): detail: Key (config_id, external_change_id)=(962f999c-a9bd-40a4-80bf-47c84b1ad750, 8e4260f3-0d56-4ea2-8510-5781642e6ef9) already exists.

adityathebe:
```
> SELECT fingerprint, jsonb_pretty(details) FROm config_changes WHERE id = 'f2650f94-b090-4f03-ab76-3e07bb107515'\G

-[ RECORD 1 ]-------------------------
fingerprint  | fe3308b7187b2370ad8e16bd8cd5bad9
jsonb_pretty | {
    "reason": "GitOperationFailed",
    "source": {
        "component": "source-controller"
    },
    "message": "failed to checkout and determine revision: unable to list remote for 'ssh://git@github.com/flanksource/aws-sandbox.git': ssh: handshake failed: knownhosts: key mismatch",
    "metadata": {
[...]

PR #897, opened 2024-09-03
fix: config_changes_config_id_external_change_id_key conflict

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/880

For a given  config, multiple changes _(distinct external change id_) could create the same fingerprint. 
Example: Two different `GitOperationFailed` events.

## Example:

| change_id | external_change_id | fingerprint | created_at |
| --------- | ------------------ | ----------- | ---------- |
| 1         | a                  | f          | 5h  |
| 2         | b                  | f          | 1m         |
| 3         | a                  | f          | 1s         |

### DIFF
--- a/changes/externalIDCache.go
+++ b/changes/externalIDCache.go
@@ -1,0 +1,33 @@
+package changes
+
+import (
+	"sync"
+
+	"github.com/flanksource/config-db/api"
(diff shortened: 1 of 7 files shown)


adityathebe:
will handle this in postgres

Issue #914, opened 2024-09-05
Intermittent duplicate key on config_items_pkey


moshloop:
failed to save results: failed to create config items: ERROR: duplicate key value violates unique constraint "config_items_pkey" (SQLSTATE 23505): detail: Key (id)=(87d8caf1-a12e-44cb-a574-dcbba7094304) already exists.

PR #916, opened 2024-09-05
fix: config_items_pkey constraint error

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/914

### DIFF
--- a/db/update.go
+++ b/db/update.go
@@ -553,7 +553,7 @@ func saveResults(ctx api.ScrapeContext, results []v1.ScrapeResult) (v1.ScrapeSum
 
 	// NOTE: On duplicate primary key do nothing
 	// because an incremental scraper might have already inserted the config item.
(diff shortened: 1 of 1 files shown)


db/update.go
```diff
@@ -551,7 +551,11 @@ func saveResults(ctx api.ScrapeContext, results []v1.ScrapeResult) (v1.ScrapeSum
 		summary.AddChangeSummary(configType, cs)
 	}
 
-	if err := ctx.DB().CreateInBatches(newConfigs, configItemsBulkInsertSize).Error; err != nil {
+	// NOTE: On duplicate primary key do nothing
+	// because an incremental scraper might have already inserted the config item.
+	if err := ctx.DB().Debug().
```
moshloop, line 556:
```suggestion
	if err := ctx.DB().
```

PR #918, opened 2024-09-06
feat: update deduped configs with postgresql func

adityathebe:


adityathebe:
we went with the trigger approach. No change is necessary on config-db.

PR #957, opened 2024-09-17
fix: update scraper id for config item

yashmehrotra:
This was causing a lot of bugs and unexpected results in catalog features

Old scrapers were replaced with new but config items' scraper id was not updated, and most of our post process jobs  use non deleted scraper_ids

### DIFF
--- a/api/cache.go
+++ b/api/cache.go
@@ -116,12 +116,8 @@ func (t *TempCache) Get(ctx ScrapeContext, id string) (*models.ConfigItem, error
 		return &item, nil
 	}
 
-	q := ctx.DB().Limit(1).Where("id = ?", id)
-	if scraperID := ctx.ScraperID(); scraperID != "" {
-		q = q.Where("scraper_id = ?", scraperID)
-	}
 	result := models.ConfigItem{}
-	if err := q.Find(&result).Error; err != nil {
+	if err := ctx.DB().Limit(1).Find(&result, "id = ? ", id).Error; err != nil {
 		return nil, dutydb.ErrorDetails(err)
 	}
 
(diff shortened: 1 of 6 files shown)


db/update.go
```diff
@@ -242,6 +242,14 @@ func updateCI(ctx api.ScrapeContext, result v1.ScrapeResult, ci, existing *model
 	if !mapStringEqual(existing.Tags, ci.Tags) {
 		updates["tags"] = ci.Tags
 	}
+
+	// This could happen when kubernetes scrapers are replaced and scrape
+	// same config items
+	if lo.FromPtr(existing.ScraperID) != lo.FromPtr(ci.ScraperID) {
+		updates["scraper_id"] = ci.ScraperID
+		ctx.Warnf("updated scraper_id of config[%s] from %s to %s", ci, existing.ScraperID, ci.ScraperID)
```
moshloop, line 250: Can we add this as a scrape warning ?

api/cache.go
```diff
@@ -117,8 +116,12 @@ func (t *TempCache) Get(ctx ScrapeContext, id string) (*models.ConfigItem, error
 		return &item, nil
 	}
 
+	q := ctx.DB().Limit(1).Where("id = ?", id)
+	if scraperID := ctx.ScraperID(); scraperID != "" {
+		q = q.Where("scraper_id = ?", scraperID)
+	}
```
moshloop, line 122: I think we need to remove this

db/update.go
yashmehrotra: You mean in job history ? or in scrape summary ?

PR #977, opened 2024-09-19
chore: do not start server on config-db run <scraper.yml>

yashmehrotra:


moshloop:
I think we do want this functionality - if you re trying to troubleshoot a slow scraper, being able to run it and then perform a trace is very valuable:

We should just make it optional: https://github.com/flanksource/mission-control/blob/main/cmd/playbook.go#L113

yashmehrotra:
@moshloop Updated

Issue #1043, opened 2024-10-08
Add config_id/type in change exclusion error


moshloop:
`error running change exclusion: failed to evaluate change exclusion expression(config_type == "Kubernetes::Pod" && jq('.metadata.ownerReferences[0].apiVersion', config.config) == "batch/v1" && ( change_type in ['Pulled', 'Started', 'Scheduled', 'Killing', 'HealthUnknown', 'Healthy'] || (change_type == 'diff' && summary.contains('metadata')) ) ): no such key: config`

PR #1045, opened 2024-10-09
fix: add details to change exclusion error

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/1043


db/update.go
```diff
@@ -274,7 +274,7 @@ func shouldExcludeChange(result *v1.ScrapeResult, changeResult v1.ChangeResult)
 
 	for _, expr := range exclusions {
 		if res, err := gomplate.RunTemplate(env, gomplate.Template{Expression: expr}); err != nil {
-			return false, fmt.Errorf("failed to evaluate change exclusion expression(%s): %w", expr, err)
+			return false, fmt.Errorf("failed to evaluate change exclusion expression (%s, expression: %s): %w", changeResult, expr, err)
```
moshloop, line 277:
```suggestion
			return false, fmt.Errorf("%s change exclusion expression failed (%s): %w", changeResult, expr, err)
```

PR #1081, opened 2024-10-16
chore: make consume events/resources use temp cache

yashmehrotra:
Fixes: https://github.com/flanksource/config-db/issues/1062


api/context.go
```diff
@@ -48,6 +50,8 @@ func (ctx ScrapeContext) withTempCache(cache *TempCache) ScrapeContext {
 	return ctx
 }
 
+var scraperTempCache = make(map[uuid.UUID]TempCache)
```
moshloop, line 53: shouldn't we use sync-map or concurrent map ?

PR #1091, opened 2024-10-18
fix: add config data to kubernetes events for post processing

yashmehrotra:


scrapers/kubernetes/kubernetes.go
```diff
@@ -241,6 +252,7 @@ func ExtractResults(ctx *KubernetesContext, objs []*unstructured.Unstructured) v
 				changeResults = append(changeResults, v1.ScrapeResult{
 					BaseScraper: ctx.config.BaseScraper,
 					Changes:     []v1.ChangeResult{*change},
+					Config:      configData,
```
adityathebe, line 255: @yashmehrotra why did we add config to changes? we have a `config == nil` check to distinguish between results that carry config item vs results with only changes.

seeing a lot of errors during extractAttributes because all changes results have an empty ID.

yashmehrotra: @adityathebe We have change transformations and exclusions which depend on config being there in cel env

Can you create an issue for the error and assign it to me ? I'll add a check to handle those

PR #1093, opened 2024-10-18
feat: hard parent for Argo

adityathebe:


db/update.go
```diff
@@ -917,7 +917,11 @@ type configExternalKey struct {
 	parentType string
 }
 
-func extractConfigsAndChangesFromResults(ctx api.ScrapeContext, scrapeStartTime time.Time, results []v1.ScrapeResult) ([]*models.ConfigItem, []*updateConfigArgs, []*models.ConfigChange, []*models.ConfigChange, map[string]v1.ChangeSummary, error) {
+func extractConfigsAndChangesFromResults(
+	ctx api.ScrapeContext,
+	scrapeStartTime time.Time,
```
yashmehrotra, line 922: I think returning a struct containing these would be simpler/readable

db/update.go
```diff
@@ -1018,14 +1022,16 @@ func extractConfigsAndChangesFromResults(ctx api.ScrapeContext, scrapeStartTime
...
 	}
-	if err := setConfigChildren(ctx, allConfigs); err != nil {
-		return nil, nil, nil, nil, allChangeSummary, fmt.Errorf("unable to set children: %w", err)
-	}
 
 	if err := setConfigPaths(ctx, allConfigs); err != nil {
 		return nil, nil, nil, nil, allChangeSummary, fmt.Errorf("unable to set config paths: %w", err)
 	}
 
+	// run this after setting the config path. else whatever the parent is set here will be overwritten by it.
```
moshloop, line 1030: Not sure I understand how this works?  child lookup hooks should default to a soft relationship, its only the argo application that shouldn't

PR #1104, opened 2024-10-23
chore: add logging rule for config change mapping

yashmehrotra:


scrapers/changes/rules.go
```diff
@@ -43,12 +44,18 @@ func (t *changeRule) process(change *v1.ChangeResult) error {
...
 		return fmt.Errorf("failed to evaluate rule (%s): %w", lo.Elipse(t.Rule, 30), err)
-	} else if !ok {
+	}
+	if !ok {
 		return nil
 	}
 
+	if properties.String("", "log.config_change_mapping_cel_env") != "" {
+		logger.Infof("result for expr=%s with env=%v is %v", t.Rule, env, ok)
+	}
```
moshloop, line 57:
```suggestion
	if ctx.PropertyOn(false, "log.rule.expr") {
		logger.Infof("result for expr=%s with env=%v is %v", t.Rule, env, ok)
	}
```

yashmehrotra: Done

Issue #1124, opened 2024-10-30
Logging improvements


moshloop:
- Only log  insights if `scraper.log.insights` is true
- Increase slow diff threshold to 1000ms

adityathebe:
> Only log insights if scraper.log.insights is true

we aren't consistently logging insights everywhere - so I'm removing the log.

PR #1136, opened 2024-11-05
chore: increase slow diff threshold to 1s from 500ms

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/1124


db/update.go
```diff
@@ -795,7 +795,8 @@ func generateConfigChange(ctx api.ScrapeContext, newConf, prev models.ConfigItem
 	if _timer != nil {
 		msg += " " + _timer.End()
 	}
-	if duration > 500*time.Millisecond {
+
+	if duration > ctx.Properties().Duration("scraper.log.slow_diff_threshold", time.Second) {
 		ctx.Logger.Warnf("SLOW DIFF >= %s", msg)
 	} else if duration > 50*time.Millisecond {
 		ctx.Logger.Infof("SLOW DIFF >= %s", msg)
```
moshloop, line 802: I think this info message can come out - we should only log warning diff times.

Issue #1146, opened 2024-11-07
Duplicate events detected 


moshloop:

moshloop:
- re-evaluate heath against original notification events
- fire health change from db

PR #1152, opened 2024-11-07
fix: health changes

adityathebe:
get the health directly from the database as a previous result in the same batch might have already updated the previous health.

**example**: 
previous health = unhealthy

result1: changes to healthy
result2: changes to healthy (❌ this should be discarded as it's already healthy)

resolves: https://github.com/flanksource/config-db/issues/1146

### DIFF
--- a/db/update.go
+++ b/db/update.go
@@ -157,7 +157,13 @@ func updateCI(ctx api.ScrapeContext, summary *v1.ScrapeSummary, result v1.Scrape
 		updates["updated_at"] = gorm.Expr("NOW()")
 	}
 
-	previousHealth := lo.CoalesceOrEmpty(lo.FromPtr(existing.Health), dutyModels.HealthUnknown)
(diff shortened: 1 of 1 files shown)


db/update.go
```diff
@@ -157,7 +157,13 @@ func updateCI(ctx api.ScrapeContext, summary *v1.ScrapeSummary, result v1.Scrape
 		updates["updated_at"] = gorm.Expr("NOW()")
 	}
 
-	previousHealth := lo.CoalesceOrEmpty(lo.FromPtr(existing.Health), dutyModels.HealthUnknown)
+	var previousHealth dutyModels.Health
+	if err := ctx.DB().Model(&models.ConfigItem{}).Where("id = ?", existing.ID).Select("health").Scan(&previousHealth).Error; err != nil {
```
moshloop, line 161: Doesn't it make more sense todo this as a trigger?

PR #1162, opened 2024-11-14
feat: trigger incremental scrape on event

adityathebe:


scrapers/event.go
```diff
@@ -0,0 +1,110 @@
...
+	// eventQueueUpdateChannel is the channel on which new events on the `event_queue` table
+	// are notified.
+	eventQueueUpdateChannel = "event_queue_updates"
+)
+
+func StartEventListener(ctx context.Context) {
+	notifyChan := make(chan string)
+	go pg.Listen(ctx, eventQueueUpdateChannel, notifyChan)
+
+	workers := 2 // TODO: decide on this
```
moshloop, line 31:
```suggestion
	workers := ctx.PropertyInt("scrapers.event.workers", 2)
```

scrapers/event.go
```diff
@@ -0,0 +1,110 @@
...
+		shutdown.ShutdownAndExit(1, fmt.Sprintf("failed to start consumer: %v", err))
+	} else {
+		go ec.Listen(ctx, notifyChan)
+	}
+}
+
+func incrementalScrapeFromEvent(ctx context.Context, event models.Event) error {
+	var configID = event.Properties["config_id"]
+
+	config, err := query.GetCachedConfig(ctx, configID)
```
moshloop, line 71: This shouldn't be a cached lookup, we should also return early if the config has been deleted or has a scrape time < 30s

scrapers/event.go
```diff
@@ -0,0 +1,121 @@
...
+		Consumer: func(_ctx context.Context, events models.Events) models.Events {
+			if len(events) == 0 {
+				return nil
+			}
+
+			if time.Since(events[0].CreatedAt) >= time.Minute {
+				// for our use case, this is considered an outdated event.
+				// we diregard it and delete it.
+				return nil
+			}
```
adityathebe, line 45: @moshloop wdyt?

PR #1175, opened 2024-11-20
fix: secret not found problem

yashmehrotra:
If we pass an object in `ctx.WithObject` it does not behave well with `ctx.GetName()` and `ctx.GetNamespace()`

Maybe we should make `.WithObject()` only take metav1.Object ?


moshloop:
I think rather log a backlog issue to fix WithObject - there is no reason it shouldn't work

Issue #441, opened 2024-03-07
K8S Connections - Native Cloud Providers 


moshloop:
i.e. specify

```
kubernetes:
  - kubeconfig:
         aws:
              cluster: 
                  value: 
              region: 
                  value: 
```

moshloop:
See https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-k8s-auth/commands/aws.go 

and this for gke via cnrm

```bash
      while read -r NAME NAMESPACE; do
                  CLUSTER=$(kubectl get containercluster $NAME -n $NAMESPACE -o yaml)
                  LOCATION=$(echo "$CLUSTER" | yq '.spec.location')
                  PROJECT=$(echo "$CLUSTER" | yq '.metadata.annotations."cnrm.cloud.google.com/project-id"')
                  TEMPFILE="$NAME-$LOCATION-$PROJECT"
                  export KUBECONFIG="$TEMPFILE"
                  export TOKEN=$(gcloud auth print-access-token)
                  gcloud container clusters get-credentials $NAME --location $LOCATION --project $PROJECT
                  unset KUBECONFIG
                  yq -i '.users[].user.token = strenv(TOKEN) | del(.users[].user.exec)' $TEMPFILE
                  kubectl create secret generic $NAME-kubeconfig -n $NAMESPACE --from-file=kubeconfig=$TEMPFILE --dry-run=client -o yaml | kubectl apply -f -
              done < <(kubectl get containercluster -A -o custom-columns=NAME:.metadata.name,NAMESPACE:.metadata.namespace | grep -v NAME)
```

PR #1195, opened 2024-11-28
feat: kubernetes aws connection

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/441


api/v1/kubernetes_connection.go
```diff
@@ -0,0 +1,128 @@
...
+
+	Cluster string `json:"cluster"`
+}
+
+type KubernetesConnection struct {
+	Kubeconfig *types.EnvVar  `json:"kubeconfig,omitempty"`
+	EKS        *EKSConnection `json:"eks,omitempty"`
+}
+
+func (t *KubernetesConnection) Populate(ctx context.Context) (kubernetes.Interface, *rest.Config, error) {
```
moshloop, line 39: this should be in duty, it would be needed for exec playbooks, kubernetesResources canaries etc..

PR #1199, opened 2024-11-29
fix: do not update created_at for deduped changes

yashmehrotra:


adityathebe:
shouldn't it be updated if the count is incremented?

yashmehrotra:
@adityathebe No, since the change already exists, updating created_at gives an incorrect indicator of when the change actually occurred first

We also have a `first_observed` column which I think should replace created at or have a new `updated_at` column

Right now, UI uses `created_at` everywhere which leads to confusion

adityathebe:
> updating created_at gives an incorrect indicator of when the change actually occurred first

The change row in the db is now representing the the creation of the latest event so `created_at` should reflect that. We already have `first_observed` to indicate when the change first occurred.

If we don't update `created_at` we won't have a "moving" dedup window.

yashmehrotra:
@adityathebe Makes sense, in that case I think we should update UI to show `first_observed` as well ?

cc: @moshloop thoughts ?

Issue #423, opened 2024-02-23
Plugins / Extensions


moshloop:
Ability to add 
- properties / relationships globally for items created by any scraper
- exclusions
- change mapping / exclusion

PR #1200, opened 2024-11-29
feat: sraper plugins

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/423


scrapers/event.go
```diff
@@ -101,6 +102,22 @@ func incrementalScrapeFromEvent(ctx context.Context, event models.Event) error {
...
+			if err := json.Unmarshal(p.Spec, &spec); err != nil {
+				return fmt.Errorf("failed to unmarshal scrape plugin spec: %w", err)
+			}
+
+			plugins = append(plugins, spec)
+		}
+	}
+
+	scrapeConfig.Spec = scrapeConfig.Spec.ApplyPlugin(plugins)
 	scrapeCtx := api.NewScrapeContext(ctx).WithScrapeConfig(&scrapeConfig)
```
moshloop, line 121: I think we can make WithScrapeConfig return a context and a scrapeconfig, and move the plugin loading there

scrapers/runscrapers.go
```diff
@@ -70,12 +71,32 @@ func RunK8IncrementalScraper(ctx api.ScrapeContext, config v1.Kubernetes, events
 
 // Run ...
 func Run(ctx api.ScrapeContext) ([]v1.ScrapeResult, error) {
+	// TODO: maybe cache this and keep it in sync with pg notify
```
moshloop, line 74: I think we might need todo that

Issue #1194, opened 2024-11-28
Add lag histograms


moshloop:
createdAt -> informed received
Informer sent -> scrape received

moshloop:
Remove - the db based lag calculation

adityathebe:
`calculateScrapeLag`

moshloop:
+ scraper_changes
+ scraper_changes_deduped (might not be very accurate due to db dedupe)

PR #1203, opened 2024-11-29
Feat/lag histogram

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/1194


scrapers/cron.go
```diff
@@ -346,13 +357,35 @@ func ConsumeKubernetesWatchJobFunc(sc api.ScrapeContext, config v1.Kubernetes) *
...
+			scrapeStart := time.Now()
+
+			if len(objs) > 0 {
+				objs = dedup(objs)
+				if err := consumeResources(ctx, scrapeConfig, config, objs); err != nil {
+					ctx.History.AddErrorf("failed to consume resources: %v", err)
+					return err
+				}
+
+				// For now, measure a single lag for the entire batch instead of per config item
```
moshloop, line 370: It needs to be per per config item, at the tail end is what is going to hurt us -

scrapers/kubernetes/informers.go
```diff
@@ -70,7 +71,15 @@ func (t *SharedInformerManager) Register(ctx api.ScrapeContext, watchResource v1
 			if ctx.Properties().On(false, "scraper.log.items") {
 				ctx.Logger.V(4).Infof("added: %s %s %s", u.GetUID(), u.GetKind(), u.GetName())
 			}
-			queue.Enqueue(u)
+
+			// TODO: We receive very old objects (months old) and that screws up the histogram
```
moshloop, line 75: Maybe we should only start recording the lag 60s after the informer is created.
The buckets should be 1s, 5s, 30s, 120s, 300s, 600s, 900s, 1800s range

scrapers/kubernetes/informers.go
```diff
@@ -89,7 +98,8 @@ func (t *SharedInformerManager) Register(ctx api.ScrapeContext, watchResource v1
 			if ctx.Properties().On(false, "scraper.log.items") {
 				ctx.Logger.V(3).Infof("updated: %s %s %s", u.GetUID(), u.GetKind(), u.GetName())
 			}
-			queue.Enqueue(u)
+
+			queue.Enqueue(NewObjectQueueItem(u))
```
moshloop, line 102: We need update lag - this is actually the most critical one - I have added an is-healthy LastUpdated func https://github.com/flanksource/is-healthy/pull/144/files#diff-b7f1feaa038882966a9d30e0b96e476dccccc2608dd8a660f80767dbbc635b6fR183

db/update.go
```diff
@@ -596,6 +596,9 @@ func saveResults(ctx api.ScrapeContext, results []v1.ScrapeResult) (v1.ScrapeSum
 	dedupWindow := ctx.Properties().Duration("changes.dedup.window", time.Hour)
 	newChanges, deduped := dedupChanges(dedupWindow, extractResult.newChanges)
 
+	ctx.Counter("config_changes", "scraper_id", ctx.ScraperID()).Add(len(newChanges))
```
moshloop, line 599: Need to add the change type and kind labels

PR #1218, opened 2024-12-04
feat: shared informer priority queue improvements

adityathebe:
* use queue for deleted objects as well. Get rid of delete buffers so every updates from the shared informers go into the queue.

### DIFF
--- a/scrapers/incremental.go
+++ b/scrapers/incremental.go
@@ -15,7 +15,6 @@ import (
 	"github.com/flanksource/config-db/scrapers/kubernetes"
 	"github.com/flanksource/config-db/utils/kube"
 	"github.com/flanksource/duty/job"
-	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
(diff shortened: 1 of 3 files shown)


scrapers/kubernetes/informers.go
```diff
@@ -332,25 +325,64 @@ func kubeConfigIdentifier(ctx api.ScrapeContext) string {
...
+	}
+
+	if opResult := pqCompareKind(qa.Obj.GetKind(), qb.Obj.GetKind()); opResult != 0 {
+		return opResult
+	}
+
+	aLastUpdatedAt := *health.GetLastUpdatedTime(qa.Obj)
+	bLastUpdatedAt := *health.GetLastUpdatedTime(qb.Obj)
+
+	if aLastUpdatedAt.Before(bLastUpdatedAt) {
```
yashmehrotra, line 354:
```suggestion
	lastUpdatedTimeA := lo.FromPtr(health.GetLastUpdatedTime(qa.Obj))
	lastUpdatedTimeB := lo.FromPtr(health.GetLastUpdatedTime(qb.Obj))

	if lastUpdatedTimeB.Before(lastUpdatedTimeA) {
		return -1
	} else if lastUpdatedTimeA.Equal(lastUpdatedTimeB) {
		return 0
	} else {
		return 1
	}

```

yashmehrotra: B before A should return -1 as that signifies A is smaller and should be processed first right ? or is it the other way around ?

scrapers/incremental.go
```diff
@@ -73,6 +73,11 @@ func ConsumeKubernetesWatchJobFunc(sc api.ScrapeContext, config v1.Kubernetes, q
 				}
 				obj := queueItem.Obj
 
+				if queueItem.Operation == "delete" {
```
yashmehrotra, line 76: (nit) lets use constant vars for this

scrapers/kubernetes/informers.go
adityathebe: > B before A should return -1 as that signifies A is smaller

`B` before `A` signifies `B` is smaller though.

PR #1219, opened 2024-12-04
run watch consumer on a shorter interval

adityathebe:


scrapers/incremental.go
```diff
@@ -31,8 +33,8 @@ func ConsumeKubernetesWatchJobFunc(sc api.ScrapeContext, config v1.Kubernetes, q
 		Context:      sc.DutyContext().WithObject(sc.ScrapeConfig().ObjectMeta),
 		JobHistory:   true,
 		Singleton:    true,
-		Retention:    job.RetentionFew,
-		Schedule:     "@every 15s",
```
yashmehrotra, line 35: Is there a reason to not keep it as an always running ?

adityathebe: I think this essentially achieves that.

we get all the other benefits that comes with a job like logging, metrics, starting/stopping based on leader election

Issue #1229, opened 2024-12-05
Metric updates


moshloop:
- scraper_deleted
- kubernetes lookup times do not seem accurate - all at  exactly 450ms

PR #1231, opened 2024-12-06
fix: metrics updates

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/1229


cmd/operator.go
```diff
@@ -118,6 +119,15 @@ func launchKopper(ctx dutyContext.Context) error {
 	return mgr.Start(ctrl.SetupSignalHandler())
 }
 
+func DeleteScrapeConfig(ctx context.Context, id string) error {
+	if err := db.DeleteScrapeConfig(ctx, id); err != nil {
+		return err
+	}
+
+	ctx.Counter("scraper_deleted", "scraper_id", id).Add(1)
```
adityathebe, line 127: not sure if we even need an id label in here.

utils/kube/fetcher.go
```diff
@@ -16,7 +16,7 @@ import (
 	v1 "github.com/flanksource/config-db/api/v1"
 )
 
-var fetchDelayBuckets = []float64{500, 1_000, 3_000, 5_000, 10_000, 20_000, 30_000, 60_000}
+var fetchDelayBuckets = []float64{10, 50, 100, 200, 300, 500, 1_000, 3_000, 5_000, 10_000, 20_000, 30_000, 60_000}
```
moshloop, line 19:
```suggestion
var fetchDelayBuckets = []float64{10, 50, 100, 500, 1_000, 5_000, 10_000, 30_000, 60_000}
```

PR #1233, opened 2024-12-06
chore: update config_change source for diff changes

yashmehrotra:


db/update.go
```diff
@@ -838,6 +838,7 @@ func generateConfigChange(ctx api.ScrapeContext, newConf, prev models.ConfigItem
 		Diff:       &diff,
 		Patches:    string(patch),
 		Summary:    strings.Join(utils.ExtractLeafNodesAndCommonParents(patchJSON), ", "),
+		Source:     v1.ChangeTypeDiffSource,
```
moshloop, line 841:
```suggestion
		Source:     v1.ChangeTypeDiff
```

yashmehrotra: Updated

db/update.go
```diff
@@ -838,6 +838,7 @@ func generateConfigChange(ctx api.ScrapeContext, newConf, prev models.ConfigItem
 		Diff:       &diff,
 		Patches:    string(patch),
 		Summary:    strings.Join(utils.ExtractLeafNodesAndCommonParents(patchJSON), ", "),
+		Source:     v1.ChangeTypeDiff,
```
adityathebe, line 841: we use "config-db" as the source for changes generated by config-db.
like health changes.
playbookStarted, ...

so let's use that probably

yashmehrotra: I intentionally didnt want that to keep the diff changes separate since their updated timestamps are handled differently

PR #1264, opened 2024-12-16
chore: remove zone relationship

yashmehrotra:
Fixes: https://github.com/flanksource/config-db/issues/1198

### DIFF
--- a/scrapers/aws/aws.go
+++ b/scrapers/aws/aws.go
@@ -122,7 +122,7 @@ func getName(tags v1.JSONStringMap, def string) string {
 }
 
 type Zone struct {
-	Region, Zone string
+	Region, Zone, ZoneID string
 }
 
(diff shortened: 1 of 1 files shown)


scrapers/aws/aws.go
```diff
@@ -1634,20 +1628,6 @@ func (aws Scraper) subnets(ctx *AWSContext, config v1.AWS, results *v1.ScrapeRes
 			Relationship:      "RegionSubnet",
 		})
 
-		// Subnet to availability zone relationship
```
moshloop, line 1637: We just need to ensure they have both the zone (us-east-1a) and id (us-east-az7) as tags

yashmehrotra: Add zone-id as tag wherever they were available, some only report zone (Volumes and LB), should I map zone -> zoneIds to populate those or is it fine ?

PR #1287, opened 2024-12-31
feat: metrics on temp cache

adityathebe:
image

huge spike in config query call in the last 2 weeks

### DIFF
--- a/api/cache.go
+++ b/api/cache.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/flanksource/commons/logger"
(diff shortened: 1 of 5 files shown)


api/cache.go
```diff
@@ -56,6 +57,12 @@ func (t *TempCache) Find(ctx ScrapeContext, lookup v1.ExternalID) (*models.Confi
 	}
 
 	if _, ok := t.notFound[lookup.Key()]; ok {
+		ctx.Counter("temp_cache_hit",
```
moshloop, line 60: I think I would use a `temp_cache_hit` and `temp_cache_missing_hit` metric names instead of using found=false

PR #1345, opened 2025-02-12
 refactor: use flat map for resource id map and global cache for dynamic gvk clients

yashmehrotra:
Merging nested maps was taking unnecessary CPU, simplified it by using a flat map instead

image


yashmehrotra:
Also, made the client caching globally because `getDelegate()` is called everytime we create a new client which is CPU Heavy

image

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Install Go
         uses: buildjet/setup-go@v5
         with:
-          go-version: 1.22.x
+          go-version: 1.23.x
```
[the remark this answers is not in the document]
yashmehrotra: You are wrong

scrapers/kubernetes/resource_map.go
```diff
@@ -1,45 +1,34 @@
...
-type ResourceIDMap map[string]map[string]map[string]string
+type ResourceIDMap map[string]string
 
 type ResourceIDMapContainer struct {
 	mu   sync.RWMutex
 	data ResourceIDMap
 }
 
+func ResourceIDMapKey(namespace, kind, name string) string {
+	return fmt.Sprintf("namespace=%s,kind=%s,name=%s", namespace, kind, name)
```
moshloop, line 18:
```suggestion
	return  namespace + kind + name
```

utils/kube/fetcher.go
```diff
@@ -36,8 +36,9 @@ func listAllCRDs(ctx api.ScrapeContext) ([]string, error) {
 	return crds, nil
 }
 
+var clientMap sync.Map
```
moshloop, line 39: I think the fetching and caching of interfaces should be in duty - especially when https://github.com/flanksource/duty/pull/1208 merges. - I am not sure how feasible but we may need to rotate the BearerToken in-place as well and/or validate the connection before taking out of the pool and refetching auth if client.ServerVersion() is not accessible

scrapers/kubernetes/resource_map.go
```diff
@@ -6,40 +6,28 @@ import (
...
-// map<namespace><kind><name>: <id>
-type ResourceIDMap map[string]map[string]map[string]string
+type ResourceIDMap map[string]string
 
 type ResourceIDMapContainer struct {
 	mu   sync.RWMutex
 	data ResourceIDMap
 }
 
+func ResourceIDMapKey(namespace, kind, name string) string {
```
adityathebe, line 16: Can we use GVK instead of just the `kind`?

yashmehrotra: That would cause a major refactor since we use GetID in a lot of places, I've created an issue to track that

utils/kube/fetcher.go
yashmehrotra: Made the change in duty: https://github.com/flanksource/duty/pull/1325

PR #1419, opened 2025-03-24
chore: add debug build

yashmehrotra:


scrapers/kubernetes/informers.go
```diff
@@ -374,20 +377,64 @@ func KindToResource(kind string) string {
...
 
-	// The object returned by the informers do not have kind and apiversion set
-	m["kind"] = resource.Kind
-	m["apiVersion"] = resource.ApiVersion
+	if properties.On(false, "log.informed_obj_size") {
+		if m != nil {
+			size := utils.MemsizeScan(&m)
+			u := &unstructured.Unstructured{Object: m}
+			if size > (100 * 1024) { // 100 KB
+				logger.Infof("Size for %s/%s/%s/%s is %s", resource.ApiVersion, resource.Kind, u.GetNamespace(), u.GetName(), HumanSize(size))
```
moshloop, line 434:
```suggestion
				logger.Debugf("Size for %s/%s/%s/%s is %s", resource.ApiVersion, resource.Kind, u.GetNamespace(), u.GetName(), HumanSize(size))
```

utils/debug.go
```diff
@@ -0,0 +1,16 @@
+package utils
```
moshloop, line 1: don't we need a memsize and a memsize_debug.go ?

yashmehrotra: we are directly overriding these in `debug.go`

Issue #1436, opened 2025-03-27
Throttle trusted advisor queries


moshloop:
i.e. scraper.aws.trusted_advisor.minInterval = 16h

PR #1437, opened 2025-03-27
feat: trusted advisor check min interval

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/1436

### DIFF
--- a/scrapers/aws/trusted_advisor.go
+++ b/scrapers/aws/trusted_advisor.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/support"
 	"github.com/flanksource/commons/logger"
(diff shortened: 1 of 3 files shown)


scrapers/aws/trusted_advisor.go
```diff
@@ -36,11 +39,41 @@ func mapSeverity(severity string) models.Severity {
...
+	query := `SELECT details->>'trusted_advisor' AS lastRan 
+	FROM job_history 
+	WHERE name = 'Scraper' 
+		AND resource_id = ? 
+		AND details->'trusted_advisor' IS NOT NULL 
+	ORDER BY time_start DESC
+	`
+
+	var lastRan string
+	if err := ctx.DB().Raw(query, ctx.ScraperID()).Scan(&lastRan).Error; err != nil {
```
moshloop, line 52: What happens if it has never run ?

PR #1461, opened 2025-04-10
fix: handle sharelock deadlock error

yashmehrotra:
Fixes: https://github.com/flanksource/config-db/issues/1449


yashmehrotra:
@moshloop Test-load failing due to another k6 dependency problem, will fix that in a separate PR

Issue #1502, opened 2025-05-20
Cloud Trail - Use principalid over role arn


moshloop:
Created by should be `john` not `Administrators`
```yaml
userIdentity:
  arn: arn:aws:sts::4324:assumed-role/Administrators/john
  type: AssumedRole
  accountId: "324"
  sessionContext:
    attributes:
      creationDate: 2025-05-16T15:59:19Z
      mfaAuthenticated: "true"
    sessionIssuer:
      arn: arn:aws:iam::21321:role/Administrators
      type: Role
      userName: Administrators
      accountId: "213213"
    webIdFederationData: {}
```

PR #1507, opened 2025-05-26
fix: cloudtrail change created by

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/1502


scrapers/aws/cloudtrail_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudtrail/types"
+	"github.com/onsi/gomega"
+	"github.com/samber/lo"
+	"gopkg.in/yaml.v3"
+)
+
+func TestCloudTrailEventToChange(t *testing.T) {
+	g := gomega.NewWithT(t)
```
moshloop, line 14: Can we add some more test cases with both assumed roles, and normal user logins

PR #1520, opened 2025-06-04
feat: gcp config access

adityathebe:


fixtures/gcp.yaml
```diff
@@ -0,0 +1,19 @@
...
+  namespace: mc
+spec:
+  gcp:
+    - project: workload-prod-eu-02
+      connection: connection://mc/gcloud-flanksource
+      include:
+        - storage.googleapis.com/Bucket
+        - container.googleapis.com/Cluster
+      auditLogs:
+        enabled: true
```
adityathebe, line 14: We don't want to enable audit logs scraping by default.

To enable, you need to use `enabled: true` instead of `include: [auditLogs]` as the default behavior of `include` is `true` when nothing is specified.

fixtures/gcp.yaml
```diff
@@ -0,0 +1,19 @@
...
+        duration: 1h
```
moshloop, line 15: This should be a max duration, and we lookup existing max on startup like we do for cloudtrail

scrapers/gcp/audit_logs.go
```diff
@@ -0,0 +1,237 @@
...
+		return "", fmt.Errorf("failed to get last saved audit log: %w", err)
+	}
+
+	filters := []string{
+		fmt.Sprintf(`logName="projects/%s/logs/cloudaudit.googleapis.com%%2Factivity"`, project),
+
+		// exclude k8s cluster audit logs from kubernetes service account
+		// because there are hundreds and thousands of these and we can't link them to any IAM Principal.
+		`(NOT (protoPayload.authenticationInfo.principalEmail:"system:" AND resource.type = "k8s_cluster"))`,
+	}
```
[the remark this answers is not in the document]
adityathebe: And we primarily need this data for "last_sign_in_time"

scrapers/gcp/audit_logs.go
```diff
@@ -100,6 +100,25 @@ func auditLogFilter(ctx *GCPContext, beginTime time.Time, project string, auditL
...
+		Where("deleted_at IS NULL").
+		Where("scraper_id = ?", ctx.ScrapeConfig().GetPersistedID().String()).Find(&externalUsers).Error; err != nil {
+		return "", fmt.Errorf("failed to get external users: %w", err)
+	}
+
+	if len(externalUsers) > 0 {
+		// Audit logs can have entries for non-IAM principals.
+		// Example: Kubernetes service accounts.
+		//
+		// We only want to fetch audit logs for the external users that we have saved in the database.
```
moshloop, line 114: I don't think we can do this - What happens if there are 1000 external users? 10000 ?

scrapers/gcp/audit_logs.go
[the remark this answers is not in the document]
moshloop: I think we should require log analytics then

scrapers/gcp/audit_logs.go
```diff
@@ -0,0 +1,257 @@
...
+	"service_account",
+}
+
+var histBuckets = []float64{1, 10, 100, 1000, 5000, 10000}
+
+const defaultAuditLogMaxDuration = 7 * 24 * time.Hour
+
+var auditLogsLastTimestampPerScraper = sync.Map{}
+
+func auditLogFilter(ctx *GCPContext, beginTime time.Time, project string, auditLogs v1.GCPAuditLogs) (string, error) {
```
moshloop, line 32: Switch to log analytics for this

PR #1522, opened 2025-06-05
fix(helm): allow additional properties for global field

yashmehrotra:


moshloop:
@yashmehrotra can you resolve conflicts ?

PR #1523, opened 2025-06-05
fix: entra matchers

adityathebe:


scrapers/azure/active_directory.go
```diff
@@ -21,13 +21,18 @@ import (
 const (
 	IncludeAuthMethods = "authMethods"
 	IncludeAppRoles    = "appRoles"
+	IncludeEntra       = "entra"
```
adityathebe, line 24: adding this include as without it it wasn't possible to just scrape entra.

PR #1526, opened 2025-06-10
feat: logs scraper

adityathebe:


scrapers/logs/logs.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return results
+}
+
+func (s LogsScraper) scrapeLoki(ctx api.ScrapeContext, config v1.Logs) (v1.ScrapeResults, error) {
+	if config.Loki == nil {
+		return nil, nil
+	}
+
+	lokiClient := loki.New(config.Loki.Loki, config.FieldMapping)
```
moshloop, line 72: Shouldn't the client crearion be pushed to duty ?

adityathebe: This is the instantiation of duty's Loki.

https://github.com/flanksource/duty/blob/main/logs/loki/loki.go#L22-L32

Issue #1534, opened 2025-06-12
GCP scraper should have option to disable asset list scraper


adityathebe:
Cuz we need a scraper that just scrapes the audit log.

adityathebe:
`failed to fetch GCP assets: error listing assets: rpc error: code = InvalidArgument desc = No RESOURCE found that matches asset type whatever for consumer project: xxxxxxxxxxx. For a list of supported content types and asset types, see https://cloud.google.com/asset-inventory/docs/supported-asset-types
`

PR #1544, opened 2025-06-16
feat: asset type vs feature flag

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/1534


api/v1/gcp.go
```diff
@@ -23,15 +24,31 @@ const (
...
+
 type GCP struct {
 	BaseScraper              `json:",inline"`
 	connection.GCPConnection `json:",inline"`
 	Project                  string `json:"project"`
 
-	// Include is a list of GCP asset types to scrape.
+	// Include specifies which GCP scraping features to enable.
+	// Supported values: "Assets", "IAMPolicy", "AuditLogs"
+	Include []string `json:"include,omitempty"`
```
moshloop, line 45: I think we should still use include for asset types and use IAMPolicy / AuditLogs as special values and/or use `auditLogs.enabled` / `iam.enabled`

PR #1572, opened 2025-06-30
feat: Health for GitHub action Workflows

adityathebe:
* fetch the runs concurrently


api/v1/github.go
```diff
@@ -3,11 +3,27 @@ package v1
...
+	// Returns workflow runs associated with a branch. Use the name of the branch of the push.
+	Actor string `yaml:"actor,omitempty" json:"actor,omitempty"`
+
+	// Returns workflow runs associated with a branch. Use the name of the branch of the push.
+	Branch string `yaml:"branch,omitempty" json:"branch,omitempty"`
+
+	// Returns workflow runs created within the given date-time range.
+	// For more information on the syntax, see "Understanding the search syntax."
+	// Docs: https://docs.github.com/en/search-github/getting-started-with-searching-on-github/understanding-the-search-syntax#query-for-dates
+	Created string `yaml:"created,omitempty" json:"created,omitempty"`
```
adityathebe, line 28: Does not seem to support relative duration. Might make sense for us to just expose: maxAge: durationString and then transform to the github date query internally.

PR #1595, opened 2025-07-14
feat: pubsub scraper

### LINKED ISSUE #1488: Pubsub / SQS/ etc Scraper

yashmehrotra:
Closes #1488


moshloop:
@yashmehrotra can you look at the other copilot comments ?

yashmehrotra:
@moshloop The problem reported is not really a problem for us, we use the key to delete scrape jobs, which happen in 2 cases:
- Scraper is deleted (in which case we unschedule all running jobs)
- Scraper is modified (in which case we unschedule all running jobs and re-run all jobs)

We used do the same for kubernetes scraper as well, since there is no advantage to storing indexes of each specific scraper since we do not unschedule on diffs

If we want to change that logic I can pick it up in a separate PR

Issue #1546, opened 2025-06-16
Better support for location hierarchy 


moshloop:
@adityathebe @yashmehrotra - What do you think of adding a logical and network location path? e.g.

aws/eu-west-1/{account | project}/cluster/namespace/
aws/eu-west-1/{account | project}/vpc/eu-west-1a/subnet/{instance}

or named locations:

deploy://github/repo/cluster/namespace/{helmrelease | argo app}
namespace://aws/eu-west-1/{account | project}/cluster/namespace/
node://aws/eu-west-1/{account | project}/vpc/eu-west-1a/subnet/{instance}
helm://{cluster}/{namespace}/{name}
kustomization://{cluster}/{namespace}/{name}
zone://
subnet://
vpc://
git://git-repo/file

_Originally posted by @moshloop in https://github.com/flanksource/mission-control/pull/2139#discussion_r2149531119_

moshloop:
```yaml

spec:

       - type: AWS::VPC
        # In config/children endpoint - for each location like alias, query {alias} in locations. for children at any level below.
          aliases:
             - network://{{tags.region}}/{{.vpcId}}
       - type: AWS::*
       # only resources that have all these tags 
       filter: all([tags.region, tags.zone, tags.vpc, tags.subnet])

#       (i.e.  parents in the right hand size details
       locations:
        - vpc://{{.tags.vp}}
        - network:{{tags.region}}/{{.tags.zone}}/{{.tags.vpc}}/{{tags.subnet}}
```

adityathebe:
https://github.com/flanksource/config-db/pull/1610

PR #1610, opened 2025-08-18
feat: config location

adityathebe:
https://github.com/flanksource/config-db/issues/1546


fixtures/plugins/k8s-locations.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: configs.flanksource.com/v1
+kind: ScrapePlugin
+metadata:
+  name: k8s-locations
```
moshloop, line 4: Shouldn't aliases and location plugins be combined ?

PR #1847, opened 2026-01-09
feat: add support for external user aliases

yashmehrotra:
## Summary by CodeRabbit

* **New Features**
  * External users can be matched by aliases for better deduplication and stable IDs across runs.
  * A background cache improves alias lookup speed and preserves IDs between scrapes.

* **Configuration**
  * Schemas/CRDs accept new fields: depth (Git) and connection (OpenSearch); OpenSearch address is now optional.

* **Tests**
  * Expanded tests cover alias matching, upsert behavior, and cache consistency across scrapes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/scrapers/run_test.go
+++ b/scrapers/run_test.go
@@ -756,9 +756,10 @@ var _ = Describe("External users e2e test", Ordered, func() {
 
 		// Verify aliases are saved
 		for _, user := range users {
-			if user.Name == "John Doe" {
+			switch user.Name {
+			case "John Doe":
 				Expect(user.Aliases).To(ContainElements("john-doe", "jdoe@example.com"))
-			} else if user.Name == "Service Bot" {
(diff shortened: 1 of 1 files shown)


db/update.go
```diff
@@ -397,6 +397,51 @@ func extractChanges(ctx api.ScrapeContext, result *v1.ScrapeResult, ci *models.C
 
 var OrphanCache = cache.New(60*time.Minute, 10*time.Minute)
 
+// ExternalUserCache stores alias+scraperID -> external_user_id mapping
+var ExternalUserCache = cache.New(time.Hour, 10*time.Minute)
```
moshloop: I don't think this would be scraper specific - multiple scrapers could be retrieving access roles and logs for the same alias

PR #2158, opened 2026-04-23
fix: persist scraper run artifacts only for explicit /run_now runs

adityathebe:
Scheduled scraper runs were writing run artifacts by default based on capture flags/config.
This made artifact persistence happen outside explicit on-demand runs

This change gates run artifact persistence behind an explicit run option, enabled only by /run_now.
Cron/scheduled runs now skip artifact writes unless explicitly triggered via /run_now, and only when artifacts.connection is configured.

## Summary by CodeRabbit

* **New Features**
  * Added a PersistRunArtifact setting to explicitly control when run artifacts are persisted.

* **Updates**
  * Artifact persistence now only occurs when PersistRunArtifact is enabled and an artifact destination is configured.
  * Run-now requests that enable any capture option will also enable artifact persistence.
  * Skipped persistence no longer records artifact counts/paths or emits skip-reason entries; persistence errors remain logged as warnings.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -162,12 +162,10 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/kms v1.49.5 // indirect
 	github.com/aws/aws-sdk-go-v2/service/signin v1.0.10 // indirect
 	github.com/aymanbagabas/go-osc52/v2 v2.0.1 // indirect
-	github.com/bahlo/generic-list-go v0.2.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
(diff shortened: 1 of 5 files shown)


moshloop:
@adityathebe  The playwright scraper is designed to capture screenshots to be saved to persistent artifact storage, so scraper artifacts are valid - the snapshots should be omitted / not even be collected during normal runs

adityathebe:
> The playwright scraper is designed to capture screenshots to be saved to persistent artifact storage, so scraper artifacts are valid

@moshloop that's the current behavior. This PR is only for conditional snapshot artifacts

PR #2301, opened 2026-07-07
feat: add CloudTrail S3 log source

adityathebe:
Add support for reading CloudTrail events from existing S3 log buckets instead of the LookupEvents API.

The S3 source lists CloudTrail date prefixes, reads gzipped log files, and converts records into the existing CloudTrail processing path for changes and AssumeRole access logs.

Progress is tracked with an in-memory S3 object cursor per bucket/region/prefix, so repeated runs in the same process skip already-processed log files without adding scrape-summary metadata.


adityathebe:
We'll use clickhouse scraper

PR #2339, opened 2026-07-31
fix(build): cross-compile Docker images instead of emulating with QEMU

adityathebe:
## Problem

Since `a33bb6c` added `platforms: linux/amd64,linux/arm64`, the release docker publish job builds the arm64 image under QEMU emulation on a single amd64 `ubuntu-latest` runner. The Dockerfile is close to a worst case for emulation — a full Rust release build (`cargo build --release` for diffgen), a cgo-linked Go build (`-tags rustdiffgen`), and `go mod download` all run emulated, which is typically 5–20× slower than native. The amd64-only baseline for "Build and push image" was ~20 minutes (run 1476); the first multi-arch run ([30620320760](https://github.com/flanksource/config-db/actions/runs/30620320760)) burned ~70 minutes on attempt 1 and was still building deep into attempt 2.

## Fix

Cross-compile instead of emulating. All builder stages are pinned to `--platform=$BUILDPLATFORM` so they always run natively on the build host, and target `$TARGETARCH`:

- **Rust**: `rustup target add aarch64-unknown-linux-gnu` + `gcc-aarch64-linux-gnu` as the linker; the staticlib is copied to `target/release/libdiffgen.a`, the fixed path `rustdiffgen.go`'s cgo directive links against.
- **Go**: `CGO_ENABLED=1 GOARCH=$TARGETARCH CC=aarch64-linux-gnu-gcc make build-prod`.
- Native builds (`BUILDARCH == TARGETARCH`, e.g. local `make docker`) skip the cross toolchain entirely and behave as before.

Only the small final runtime stage still runs under emulation (`useradd` + `config-db go-offline`, which is network-bound).

Also included:

- `cargo fetch` now depends only on `Cargo.toml`/`Cargo.lock` (with a stub `lib.rs`), so the dependency layer survives diffgen source changes.
- The go builder copies only `target/release` from the rust stage instead of the entire multi-GB `target` directory.
- The PR **Build** workflow now builds both release platforms with buildx (timeout 20 → 30 min), so cross-compilation breakage is caught on PRs instead of on release. It reuses the same pinned `docker/setup-buildx-action` SHA as `flanksource/action-workflows`.

## Validation

- Cross-compiled `diffgen` to `aarch64-unknown-linux-gnu` with the exact toolchain packages from the Dockerfile; verified the produced staticlib/cdylib are `ELF 64-bit ARM aarch64`.
- Cross-compiled the full config-db binary with `CGO_ENABLED=1 GOARCH=arm64 CC=aarch64-linux-gnu-gcc make build-prod` linking the cross-built `libdiffgen.a` — build succeeded and the output is `ELF 64-bit LSB executable, ARM aarch64, dynamically linked, interpreter /lib/ld-linux-aarch64.so.1`.
- Verified the pinned `flanksource/base-image:0.7.1` digest is a multi-arch index containing amd64 and arm64.
- Exercised the Dockerfile's `case`/`if` shell logic for amd64-native, arm64-cross, and unsupported-arch paths.
- This PR's updated Build workflow additionally runs the full multi-arch image build end-to-end as CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016NaE7LcgYZXbx5igVaN2Rg

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ permissions:
 jobs:
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
     steps:
       - name: Harden Runner
         uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
@@ -30,7 +30,12 @@ jobs:
           difference=$((final_space - initial_space))
           echo "Disk space difference (in KB): $difference"
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
+
+      # Build both release platforms so PRs catch cross-compilation breakage
+      # before it reaches the release workflow
       - name: Build Container
-        run: make docker
+        run: docker buildx build --platform linux/amd64,linux/arm64 -f build/Dockerfile .
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,17 +1,40 @@
-FROM rust:bookworm@sha256:29fe4376919e25b7587a1063d7b521d9db735fc137d3cf30ae41eb326d209471 AS rust-builder-base
-COPY external/diffgen/ /app/external/diffgen
-RUN cd /app/external/diffgen/ && cargo fetch
+## Builder stages run on the build platform and cross-compile to TARGETARCH,
+## so multi-arch builds don't pay for QEMU emulation of cargo/rustc/go.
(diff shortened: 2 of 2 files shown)


adityathebe:
CI note: `test-clickhouse` is red on this PR, but the failure is pre-existing — the identical failure occurred on #2338 ([job](https://github.com/flanksource/config-db/actions/runs/30619050706/job/91118963472)) before this branch existed. The job fails during `helm install` in the kind cluster: `no matches for kind "ScrapePlugin" in version "configs.flanksource.com/v1" ... ensure CRDs are installed first` — the chart's `ScrapePlugin` resources are applied before that CRD is registered. This PR only touches `build/Dockerfile` and the Build workflow, so I'm not fixing it here; it needs a chart/CRD-ordering fix (e.g. moving the ScrapePlugin CRD into the chart's `crds/` directory or gating the plugin resources).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2359, opened 2026-08-17
fix: resolve relationship external IDs

adityathebe:
## Summary

- normalize `RelationshipSelector.ExternalID` and query the dedicated `config_items.external_id` array
- retain all other relationship selector constraints before filtering by external ID
- add regression coverage for matching and mismatched selector types

## Tests

- `go test ./db -run '^TestDB$' -count=1 -ginkgo.focus 'FindConfigsByRelationshipSelector'`\n- `go test ./db -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relationship selector matching for configurations using external IDs.
  * Selectors combining external IDs with other criteria now return only matching configurations.
  * Selectors with unmatched external IDs correctly return no results.

### DIFF
--- a/db/config.go
+++ b/db/config.go
@@ -56,14 +56,41 @@ func FindConfigIDsByRelationshipSelector(ctx context.Context, selector duty.Rela
 		return nil, nil
 	}
 
-	return query.FindConfigIDsByResourceSelector(ctx, 0, selector.ToResourceSelector())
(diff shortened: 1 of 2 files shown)


adityathebe:
@coderabbitai review

adityathebe:
Duty is probably the better place for this fix.

PR #2380, opened 2026-09-04
chore(gcp): update cost-resource mapping

yashmehrotra:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google Cloud cost matching by prioritizing full resource names, reducing incorrect matches when resource identifiers share similar final segments.
  * Added a fallback to the final resource-name segment when no exact full-name match is found.
  * Preserved root configuration handling for ambiguous exact matches.


yashmehrotra:
@coderabbitai review

